PR #5, opened 2026-09-09
perf(auth): verified bearer tokens are remembered, so bcrypt leaves the ingest hot path

nidhi-nair:
## Why

`ApiKeyService.verify` runs on every authenticated request and did the same three things each time for a credential that never changes: a `findByPrefix` SELECT, a **bcrypt verification at cost 10**, and an `UPDATE api_key SET last_used_at`.

A JFR profile of the OTLP ingest path attributed **62% of all backend CPU** to `BCryptOpenBSDProtocol.cryptRaw`, against **1.1%** for actually writing the spans and 0.1% for protobuf decode. The `last_used_at` write made every push from one exporter an `UPDATE` of the *same single row* — 61,845 updates against 1 live row over a benchmark run, with the row-lock contention and dead-tuple churn that implies.

## What changed

`VerifiedTokenCache` keeps the verification result for a short while. The bounds are the point, so each abuse case is handled explicitly:

- **Verified and rejected tokens live in separate bounded maps.** A single map would be the real hole: a flood of invented tokens would evict every genuine entry and put bcrypt back on the hot path exactly under load. Only a caller holding a real token can displace a verified entry.
- **Both evict least-recently-used and never refuse an insert** — a cache that refuses is one an attacker can use to lock out the people it was built for. Overflow costs a bcrypt, which is where we started.
- **`revoke()` and `rotate()` evict the key inside the same call that writes the revocation**, so a key withdrawn through the product stops verifying immediately. The TTL is a backstop for changes that never went through those methods (a hand-edited row, a restored backup, another replica), not the revocation control.
- **Entries are keyed on the SHA-256 of the token, never the token**, so a heap dump yields digests rather than working credentials.
- **`evals.auth.token-cache.enabled=false` restores the previous behaviour exactly** — and the A/B below uses that switch, which tests the claim.

`last_used_at` is now written at most once per key per interval (default 60s). The column becomes accurate to the minute; `audit_log` is untouched.

## Measured

A/B on the self-host reference stack (4 vCPU / 8 GiB; backend 2 vCPU / 2 GiB, Postgres 1.25 vCPU / 3 GiB), **identical database state — substrate truncated before each arm**, with the cache flag as the only variable. 120s per case after a discarded warm-up.

| Case (800 spans/s offered unless noted) | Accepted, no cache | Accepted, cached | p99 before → after |
|---|---|---|---|
| 40 spans/push, 400 offered | 373.7 /s | 370.7 /s | 67 ms → **5.9 ms** |
| 40 spans/push, saturated | 416.8 /s | **475.8 /s** | 141 ms → **6 ms** |
| **10 spans/push**, saturated | 12.9 /s | **421.0 /s** | 38.1 s → **28 ms** |
| 200 spans/push, saturated | 478.1 /s | 474.3 /s | 77 ms → 37 ms |

**Read the 200-spans row first: it does not move.** That case was already bound by the single drainer thread and Postgres, not by auth. This change does not raise the ceiling; it removes a cost that was stopping most workloads from reaching it.

What actually changed:

- **Capacity stopped depending on the customer's exporter.** Accepted throughput ranged 13–478 spans/s purely on batch size before; 421–476 after. Flat.
- **Non-retryable 500s went to zero.** The no-cache small-batch case produced 34 `CannotGetJdbcConnectionException` 500s; the cached arm produced none. OTLP clients retry 503 but not 500, so those were silent span loss.
- **Writes to `api_key`: 8,549 → 15** across the same workload.
- **The bottleneck moved out of our code.** Backend now ~70% CPU while Postgres pins at ~126% of its 1.25 vCPU.

Validation that the flag is a true control: the no-cache arm returns p50 to **59 ms**, against the **61 ms** measured before this cache existed.

## Notes for review

- **No tests added**, per `AGENTS.md` ("Do not add or modify tests unless explicitly requested"). For a security-relevant cache I would want three: separate-cap eviction, revocation-invalidates-immediately, and negative-TTL expiry. Happy to add them if the team wants them here rather than as a follow-up. The existing `ApiKeyServiceTest` / `ApiKeyManagementTest` already assert revoke and rotate take effect immediately, and both pass.
- `scripts/check-backend.sh` green: **590 tests, 0 failures**, including `ArchitectureRulesTest`.
- Config keys documented in `devdocs/reference/config-keys.md` in this PR.
- **Not fixed here:** `DataSourceConfig` hardcodes `setMaximumPoolSize(10)`, so the whole `spring.datasource.hikari.*` block in `application-production.yaml` is dead config (`connection-timeout: 10000` never applies; Hikari's 30s default is what the measured 34s p99 reflected). This change removed the pressure, not the dead config. Separate PR.


nidhi-nair:
## Code Review Summary

Found 2 blocker(s) and 4 warning(s). **Verdict: changes-requested.**

Reviewed as a substantial/risky diff (auth + concurrency), so the full advisory panel was convened
(architect, pragmatist, perf-analyst, visionary, product-advocate) and their findings deduped below.
Benchmarks were not re-run; every claim below was verified against the code.

The design is good and the reasoning in the class docs is mostly right. Two things are wrong: one
revocation path never reaches the cache, and one does but can be overtaken.

---

### Blockers

**1. (Security) Project deletion revokes keys without invalidating the cache — cached keys keep ingesting for up to `ttl-seconds`**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/ApiKeyService.java:204` is the only place a
revocation reaches the cache, and it is reached only from `ApiKeyService.revoke`/`rotate`.
`TenantService.deleteProjectAsync` (`backend/tenancy/src/main/java/ai/tessary/evals/tenant/TenantService.java:410`)
revokes a project's whole key set through `ApiKeyRepository.revokeAllForProject`, straight at the
repository, and never touches `VerifiedTokenCache`. Any key of that project verified in the previous
`ttl-seconds` (default 60) keeps authenticating for the rest of its TTL.

This is not a theoretical gap — it is the exact window `ApiKeyRepository.revokeAllForProject`'s own
javadoc (`ApiKeyRepository.java:89-97`) says must not exist:

> This is the whole access story for a project being deleted, and it is deliberately the part that
> does NOT run in the background. […] revoking here means nothing new lands in a project on its way
> out — whereas a revocation queued behind the purge would leave a window, minutes wide, in which the
> deleted project is still accepting writes.

And it bites hardest on exactly the traffic this PR optimises. Project-scoped REST is still gated
(`auth/TenantPathResolver.java:65` refuses a project marked `deleting`), but **the ingest path is
not**: `OtlpGrpcAuthInterceptor` resolves purely through `BearerTokenAuthenticator`, and
`BearerTokenAuthenticator.authenticate` only checks that the project row *exists* — a deleting
project still has its row. `OtlpTraceController` has no `isDeleting` check either. So a write-scoped
key cached before the delete keeps accepting spans into a doomed project for up to a minute.

Also note `evals.auth.token-cache.enabled=false` cannot rescue this path: it never consults the
cache at all, so the flag only makes the cache inert everywhere else.

Fix (minimal): give `VerifiedTokenCache` a project-scoped evict and call it from
`deleteProjectAsync` in the same transaction —

```java
public void invalidateProject(String projectId) {
    synchronized (this) {
        verified.entrySet().removeIf(e -> e.getValue().key().projectId().equals(projectId));
    }
}
```

`invalidateAll()` from `deleteProjectAsync` also closes it and is one line (and would give that
currently-uncalled method a caller — see warning 3), at the cost of flushing unrelated tenants'
entries on every project delete.

**2. (Security) `invalidate()` can be overtaken by a verification already in flight, re-caching a revoked key for a full TTL**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/ApiKeyService.java:182` writes the cache entry
*after* the bcrypt, with no check that the row is still the row it read. Interleaving:

1. Request thread: `verify()` misses the cache, `findByPrefix` reads a live row, enters
   `BCrypt.verifyer().verify(...)` — tens of ms, and materially longer under the load this PR
   measured.
2. Admin thread: `revoke()` writes `revoked_at`, then `cache.invalidate(tokenId)` — which finds
   nothing, because the request thread has not inserted its entry yet.
3. Request thread returns from bcrypt, `result.verified == true`, and calls
   `cache.rememberVerified(presented, row.get())` with the **stale, unrevoked** row.

The revoked key now verifies from cache for the full `ttl-seconds`. This contradicts the property
the class doc states as its central safety claim (`VerifiedTokenCache.java:33-40`, "a key withdrawn
through the product stops verifying immediately rather than at the end of a TTL") and the same claim
in the PR body.

The window is one bcrypt per token per TTL, so a randomly-timed revoke hits it rarely — but the
scenario that motivates a revoke is a key under active abuse, the consequence is a 60s
authentication bypass on the credential you just withdrew, and the fix is small.

Fix (minimal): a generation counter, which also covers blocker 1 for free.

```java
private long generation;                                  // guarded by the same monitor

public long generation() { synchronized (this) { return generation; } }

// in every invalidate*(): generation++;

public void rememberVerified(String presented, ApiKey key, long seenAtGeneration) {
    if (!props.isEnabled()) return;
    long now = System.currentTimeMillis();
    Entry e = new Entry(key, now + props.getTtlSeconds() * 1000L, now);
    synchronized (this) {
        if (generation != seenAtGeneration) return;       // a revocation raced us; do not cache
        verified.put(digest(presented), e);
    }
}
```

with `verify()` capturing `long gen = cache.generation();` before `tokens.findByPrefix(prefix)`.

**Panel divergence, flagged for the author/maintainer to settle:** the perf-analyst reviewed this
same path and returned "verified correct — revocation is not TTL-bound", having considered
`invalidate()` against `lookup()` (which is indeed safe) but not against a `rememberVerified()`
already in flight. I am keeping it as a blocker because it is an auth-bypass window on the one
property the class exists to guarantee, and the fix is ~6 lines. If you disagree that the race is
worth closing, the alternative is to soften the absolute claim in `VerifiedTokenCache.java:33-40`,
`TokenCacheProperties.java:33-40` and `config-keys.md:47` — but shipping the absolute claim with the
race in place is not an option.

---

### Warnings

**3. (Tests) The no-tests decision was correct to make, and I am now explicitly requesting three of them**

`AGENTS.md` says "Do not add or modify tests unless explicitly requested", and the PR was right to
follow it and to flag the tension rather than quietly break the rule. But `backend/AGENTS.md:151`
puts "token verification (revoked rejection, `last_used_at`)" on the *What to test* list, and this
change moves that verification onto a new caching path. More concretely: the existing suite is not
evidence of safety here. `ApiKeyServiceTest.verify_rejectsRevokedToken` (`:75-78`) and
`ApiKeyManagementTest.rotate_revokesOld_issuesFreshWithSameScope` (`:70-80`) do genuinely warm the
cache before revoking — the PR body's claim about them is accurate — but the path that is actually
broken (blocker 1) has no test at all, and passed green.

Treat this as the explicit request the rule asks for. Required before merge:

1. `verify_rejectsKeyAfterProjectDeletion` — issue a key, `verify()` it to warm the cache, call
   `TenantService.deleteProjectAsync`, assert `verify()` is empty. This fails today.
2. `verify_rejectsRevokedKeyCachedByAnInFlightVerification` — the blocker-2 race. Deterministic
   version: capture the generation, revoke, then call `rememberVerified` with the stale generation
   and assert the entry is not cached.
3. `verify_floodOfRejectedTokensDoesNotEvictVerifiedEntries` — insert one verified entry, drive
   `max-rejections * 4` distinct invalid tokens through `verify()`, assert the verified entry is
   still served from cache. This is the load-bearing claim of the two-map design and nothing pins it.

A fourth would be welcome but is not a gate: `verify_expiresAfterTtl` with
`ttl-seconds` driven from `@DynamicPropertySource`.

**4. (Consistency) `invalidateAll()` has no caller anywhere in the tree**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/VerifiedTokenCache.java:164`. Its javadoc
promises "for tests and for an operator who wants the cache cold without a restart", but there is no
test (see warning 3) and no operator surface — no controller, no actuator endpoint. Against
`AGENTS.md`'s "NEVER keep old implementation as fallback. Delete dead code", this ships as an
unreachable public method advertising a capability that does not exist. Either give it the caller
blocker 1 needs, or drop it and the "operator" half of the sentence.

**5. (Docs) The same threat model is written out three times, in full**

`VerifiedTokenCache.java:33-40`, `TokenCacheProperties.java:33-40` and
`devdocs/reference/config-keys.md:47` each independently derive "revoke/rotate evict synchronously,
the TTL is a backstop for a hand-edited row / a restored backup / another replica", in near-identical
words. `AGENTS.md`: "One home per fact. Never duplicate a rule across files — restate at most one
line and link to the owning doc." This matters more than usual here because the fact being triplicated
is a security invariant that blockers 1 and 2 both change: a fix now has to be remembered in three
places. Keep the full version on `VerifiedTokenCache` (the class that enforces it); reduce the other
two to one line plus a pointer. The `last-used-write-interval-seconds` rationale and the 61,845-updates
measurement are duplicated the same way between `TokenCacheProperties.java:66-76` and `config-keys.md:47`.

**6. (Docs) `devdocs/reference/auth-and-mcp.md` was not co-updated**

That file is the factual inventory of the auth surface and its class table (`:125-135`) lists
`tenant/ApiKeyService` as "Issue / verify / revoke (bcrypt at rest, prefix lookup)". After this PR
that line is incomplete and `tenant/VerifiedTokenCache` is missing from a table an operator reads to
understand how bearer auth resolves — including the one fact they most need, that a revocation
performed outside `ApiKeyService` is not immediate. `AGENTS.md` requires the same-PR co-update.
`devdocs/reference/architecture.md:352` is package-level and needs no change.

---

### Suggestions

- `backend/tenancy/src/main/java/ai/tessary/evals/tenant/TokenCacheProperties.java:90`, `:109`, `:128` —
  `maxEntries`/`maxRejections` validate `< 1` and throw; `ttlSeconds`, `negativeTtlSeconds` and
  `lastUsedWriteIntervalSeconds` do not. Traced all of them: every out-of-range value degrades
  *towards* disabling the optimisation, never towards fail-open, so this is not a security issue.
  But `ttl-seconds: 0` silently makes a cache that reports `enabled: true` never return a hit, which
  is a confusing thing to debug. Either add the same guard, or state in the javadoc that `<= 0`
  disables. (`negative-ttl-seconds <= 0` is already handled explicitly and correctly at
  `VerifiedTokenCache.java:143`.)
- `backend/tenancy/src/main/java/ai/tessary/evals/tenant/VerifiedTokenCache.java:111` —
  `synchronized (this)` on a Spring singleton publishes the lock to anything holding the bean. A
  `private final Object lock` is the same code and cannot be contended from outside. Optionally one
  lock per map, which would stop an invented-token flood sharing a monitor with genuine
  verifications; at today's volumes that coupling is not measurable, so this is preference, not perf.
- `VerifiedTokenCache.java:171-178` — a fresh `MessageDigest` per call, up to twice per request.
  Negligible against the bcrypt removed; noting it only so it is a known cost rather than a
  discovered one.
- The A/B harness the *Measured* table rests on is `docker-compose.bench.yml`, which is untracked and
  not part of this PR — the numbers cannot be reproduced from the repo as it will land. Worth
  checking it in with the change that justifies itself by them.
- `last_used_at` staleness is user-visible, mildly: `frontend/src/views/Settings/ApiKeys.tsx:389` and
  `McpTokens.tsx:321` render `relativeTime(last_used_at)`, so a key in continuous use can show
  "Just now" for up to a minute after it stopped being used, and a user checking whether a key is
  idle before revoking it sees a value up to 60s stale. No contract promises freshness and the UI
  already rounds to whole minutes, so this is fine — but the PR frames the change as purely a
  DB-write concern and it is worth one sentence acknowledging the UI effect.

---

### Verified, and clean

Stated explicitly because they were the questions asked of this review, and because each survived a
real check rather than a glance:

- **Two-map separation holds.** `rememberVerified` is reachable only after
  `BCrypt.verifyer().verify` succeeds against a live row (`ApiKeyService.java:177-182`). A flood of
  invented tokens can only drive `rejected`, and evicts nothing but other rejections. The design's
  central claim is true.
- **`removeEldestEntry` is right on every insertion path.** It is consulted by `put`/`putAll`, never
  `get`. The refresh re-put at `VerifiedTokenCache.java:117` replaces an existing key, so
  `HashMap.putVal` calls `afterNodeAccess` but not `afterNodeInsertion` — no double-count, no leak.
  Both maps stay bounded.
- **No virtual-thread pinning.** `backend/pom.xml:41` is Java 25; JEP 491 made `synchronized`
  non-pinning permanently in JDK 24, and there is no `Object.wait` or native frame in the critical
  section. The class doc's claim is correct on this JDK.
- **SHA-256 keying, and nothing sensitive newly retained.** Keys are Base64(SHA-256(token))
  (`VerifiedTokenCache.java:171-178`); a token is 132 bits of randomness, so the digest is not
  invertible. The cached `ApiKey` carries the bcrypt hash the DB already returns per request — held
  for 60s rather than one request, which is not a new class of secret. `MessageDigest.isEqual` is
  correctly judged unnecessary here.
- **`enabled=false` genuinely restores the old behaviour** on `verify` — `lookup` returns `Unknown`
  unconditionally, both remember methods no-op, and `markUsed` then runs on every request exactly as
  before (`ApiKeyService.java:183`). The one path it does *not* restore is blocker 1, which never
  consults the cache in either state.
- **Nothing depends on per-request `last_used_at` freshness.** The only readers are the two Settings
  views above, via `ApiKeyController.java:146` / `McpTokenController.java:104`. No rate limiter, no
  expiry check, no doc or OpenAPI schema promises freshness. `audit_log` is genuinely untouched.
- **PR-body claims (a)–(e)** all check out against the code, with the single exception noted in
  blocker 1: (a) is true of `revoke`/`rotate` but not of bulk project revocation.
- **No `guardrails.protected_paths` touched** — no migrations, Terraform, Dockerfile, `.env` or
  `.github/**` in the diff.

---

`needs-human` applied: this is an authentication path, blocker 2 is a split call the panel did not
reach agreement on, and crew does not decide a merge. A human owns the call on both blockers.

🤖 crew review

nidhi-nair:
Both blockers fixed in 3f21782. I verified each one against the code before fixing rather than taking the review at face value — both were real.

**Blocker 1 (project deletion) — confirmed and fixed.** `ApiKeyRepository.revokeAllForProject`'s own javadoc is the indictment: it calls itself *"the whole access story for a project being deleted"*, deliberately synchronous so that *"nothing new lands in a project on its way out — whereas a revocation queued behind the purge would leave a window, minutes wide"*. The cache handed that window straight back, 60s wide, on the one path that authenticates on the key alone. `TenantService.deleteProjectAsync` now calls `invalidateProject` in the same method that writes the bulk revoke.

**Blocker 2 (in-flight race) — confirmed and fixed.** The reviewer was right and the perf-analyst's dissent was checking the wrong pair: `invalidate()` vs `lookup()` is indeed safe, but `invalidate()` vs an in-flight `rememberVerified()` is not. Every invalidation now bumps a generation counter; `verify` reads it *before* the row read and `rememberVerified` re-checks it under the monitor before storing, so a result derived from a pre-revocation read is dropped. Bumping inside the monitor and before the sweep makes it correct whether the racing store would have landed before or after the removal.

**On the tests.** The review is the explicit request `AGENTS.md` asks for, so five are added. The one that matters: `projectDeletion_stopsCachedKeysVerifying` was run with the one-line fix reverted and **fails** (`expected: <true> but was: <false>`), then passes with it restored. It holds the bypass shut rather than exercising the method. Also covered: the invalidation race, the flood-does-not-evict-verified-entries claim that is the whole justification for two maps, the kill switch, and revoke.

**Also addressed:** `invalidateAll()` had no caller and is deleted; the threat model is stated once on the class instead of three times; `devdocs/reference/auth-and-mcp.md` now carries the invariant that *any* new bulk-revocation path must invalidate too — which is the general form of blocker 1 and the thing most likely to be re-broken.

**Not addressed, deliberately:** the bench harness is still untracked, so the *Measured* table is not reproducible from the repo as this lands. I'd rather add it as its own change than expand this PR. Flagging it so the claim is read with that caveat.

Gate: `scripts/check-backend.sh` — **595 tests, 0 failures**.

Separately, and worth knowing before anyone reads the perf numbers as a ceiling: I measured what the next bottleneck is. Giving Postgres 2.4× the CPU (1.25 → 3.0 vCPU) moves throughput **+1.5%** — it peaks at 127% of 300% available and still goes no faster, while the backend sits at 96–99%, one saturated core. The wall is the **single drainer thread** (`IngestSpool.drainers()` returns 1), not the database.

nidhi-nair:
## Code Review Summary

Re-review of `3f21782` against the prior review, read as an incremental diff (`8db9a13..HEAD`) in the
context of the full change. Substantial/risky diff (auth + concurrency), so the advisory panel was
convened again and its findings deduped below. Every claim was verified against the code; no
benchmarks re-run.

**Blocker 2 (in-flight re-cache) is closed.** **Blocker 1 (project deletion) is closed for the case
it was reported as, and reopened in a narrower form by where the new call sits.** Found 1 blocker
and 4 warnings.

---

### Blockers

**1. (Security) `invalidateProject` is called *inside* the still-open transaction, so the generation bump lands before the revoke is visible to any other connection**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/TenantService.java:407-419`. `deleteProjectAsync`
is `@Transactional`, and the new line sits between the UPDATE and the commit:

```java
@Transactional
public ProjectDeleteAcceptance deleteProjectAsync(String projectId, String now) {
    ...
    int revoked = apiKeys.revokeAllForProject(projectId, now);   // UPDATE issued, NOT yet committed
    tokenCache.invalidateProject(projectId);                     // generation bump + sweep happen HERE
    deleteJobs.enqueue(projectId, now);
    return ...;
}                                                                // commit happens on return
```

Neither `ApiKeyService` nor `ApiKeyRepository` is `@Transactional`, so a concurrent verification runs
on its own autocommit connection under READ COMMITTED and cannot see the UPDATE until this method
returns. That gives an interleaving the generation counter was designed to catch but cannot:

1. Thread A: `revokeAllForProject` — uncommitted.
2. Thread A: `invalidateProject` — sweeps the map and bumps `generation` to *n+1*
   (`VerifiedTokenCache.java:206-216`).
3. Thread B (ingest, cache miss): `observedGeneration = cache.generation()`
   (`ApiKeyService.java:172`) reads the **already-bumped** *n+1*.
4. Thread B: `findByPrefix` (`ApiKeyService.java:174`) reads the pre-commit row — `revoked_at` still
   `NULL`. Live key.
5. Thread B: bcrypt passes (revocation does not touch the hash).
6. Thread B: `rememberVerified(..., n+1)` (`ApiKeyService.java:185`) — generation still *n+1*, no
   invalidation has landed since step 3, so the entry **is stored**.
7. Thread A commits.

A key of a project whose deletion has been accepted now authenticates ingest for a full
`ttl-seconds` (60s default). That is the same bypass, on the same path, that this commit exists to
close — and it is specifically the path the class doc calls "the one that matters most"
(`VerifiedTokenCache.java:41-47`). It is real on the ingest path because
`BearerTokenAuthenticator.authenticate` only checks the project row *exists* and
`ProjectRepository.findById:20-25` does not filter `deleting_at`; the `isDeleting` gate lives only in
`TenantPathResolver:65,92`, which OTLP does not traverse.

The single-key path does **not** have this problem, and the contrast is the proof: `ApiKeyService`
is not `@Transactional`, so `tokens.revoke(...)` autocommits before `cache.invalidate(...)` runs
(`ApiKeyService.java:110-111`, `:206-207`). The generation mechanism's contract — "the invalidation
must not become visible before the revoke does" — holds everywhere except the one call this commit
added.

Neither new test can catch it: `invalidationDuringVerification_isNotOverwritten`
(`VerifiedTokenCacheTest.java:71-88`) drives the three cache methods sequentially on one thread and
never goes through `deleteProjectAsync`, and `projectDeletion_stopsCachedKeysVerifying` (`:48-68`)
calls `verify()` only after `deleteProjectAsync` has returned, i.e. after commit.

**Fix (minimal), and the pattern already exists in this file:** make the eviction a post-commit
action rather than an in-transaction one. `TenantService.java:330-334` already publishes
`ProjectCreatedEvent` `AFTER_COMMIT` for exactly this reason ("so a listener never sees a project
that rolled back"). Either register a synchronization —

```java
TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
    @Override public void afterCompletion(int status) { tokenCache.invalidateProject(projectId); }
});
```

(`afterCompletion` rather than `afterCommit` so a rollback also clears — evicting after a rollback
is fail-safe, it only costs a bcrypt) — or move the call to an `@TransactionalEventListener(phase =
AFTER_COMMIT)`. Post-commit ordering makes the generation check correct again: a verifier that read
the old generation is refused at `rememberVerified`, and a verifier that reads the new generation
necessarily reads the row after the commit.

**Panel note, flagged for the author to settle:** the perf-analyst returned this as a blocker; the
pragmatist returned it as "warning, bordering blocker" on the grounds that the window is the commit
round-trip (milliseconds) and needs a concurrent cache-*miss* on that project to land inside it. Both
agree the bug is real and the fix is ~5 lines. I am keeping it a blocker for consistency with the
standard the last review set — blocker 2 was accepted as a blocker on a comparably narrow window,
for the same property, with a comparably small fix — but the probability argument is honest and a
human owns the call.

---

### Warnings

**2. (Docs) The commit message says the threat model is now "stated once on the class rather than three times". It is not — it is now stated four times, and two of the copies are stale in exactly the way last review predicted**

`3f21782` touched `VerifiedTokenCache.java` and `auth-and-mcp.md` but not `TokenCacheProperties.java`
or `config-keys.md`. Both still carry the full derivation, and both now say something false by
omission:

- `backend/tenancy/src/main/java/ai/tessary/evals/tenant/TokenCacheProperties.java:35-41` — "This is
  *not* the revocation window: `ApiKeyService#revoke` and `#rotate` evict the key's entry as part of
  the same call... The TTL bounds the cases that do not go through those methods — a row deleted by
  hand in psql, a restore from backup, a second replica."
- `devdocs/reference/config-keys.md:47` — "`ttl-seconds` (60) is a backstop only — `ApiKeyService.revoke`/`rotate`
  evict the entry in the same call... the TTL covers changes that never went through those methods."

Project deletion is a third synchronous evictor and neither mentions it; read literally, both tell a
reader that bulk project revocation is TTL-bound. This is the failure mode warning 5 named — a
security invariant triplicated is one a fix has to be remembered in three places — and it happened on
the very next commit. Reduce both to one line plus a pointer to `VerifiedTokenCache`, which is where
the full version now correctly lives.

**3. (Docs) `auth-and-mcp.md`'s code map still omits the class, so warning 6 is only half addressed**

The new prose at `devdocs/reference/auth-and-mcp.md:42-54` is accurate and is the right content — it
names all three guarantees and the "any new bulk-revocation path has to invalidate too" invariant.
But the **Code map** table at `:144` still reads `tenant/ApiKeyService` | "Issue / verify / revoke
(bcrypt at rest, prefix lookup)" and there is still no `tenant/VerifiedTokenCache` row. That table is
what an operator scans; a caching layer on the bearer-auth path should appear in it. One row plus
"verify is cache-backed" on the `ApiKeyService` line closes it.

**4. (Tests) The tests are real — three of them genuinely fail with the bug present — but two claims in them are not actually asserted, and the one ordering the fix depends on is unpinned**

Confirmed load-bearing, not tautological: `projectDeletion_stopsCachedKeysVerifying` (`:48-68`) fails
without `invalidateProject`; `floodOfInvalidTokens_doesNotEvictVerifiedEntries` (`:94-113`) fails with
a single map; `invalidationDuringVerification_isNotOverwritten` (`:71-88`) fails without the
generation guard. Good tests. Three gaps:

- `backend/app/src/test/java/ai/tessary/evals/tenant/VerifiedTokenCacheTest.java:143-147` —
  `assertEquals(0, tokens.verify(...).stream().count())` is a restatement of the `isEmpty()`
  assertion two lines above it, in a more roundabout form (`Optional.stream().count() == 0` **iff**
  empty). It is not a tautology — the second `verify()` call does exercise the negative-cache path —
  but it asserts nothing about the thing its message names ("served from the rejection memory"), and
  the method javadoc's claim that "the rejection must not be cached as a false positive" is not
  tested at all. Assert the mechanism instead: `assertInstanceOf(Lookup.Rejected.class,
  cache.lookup(issued.plaintext()))`.
- Nothing pins the ordering the whole generation mechanism rests on. Move
  `long observedGeneration = cache.generation();` (`ApiKeyService.java:172`) below the
  `findByPrefix` on line 174 and every one of these five tests still passes, while the mechanism
  becomes a no-op. That ordering is a comment today; it should be a test (or the two statements
  should be fused so they cannot be separated).
- Negative-TTL expiry — the third of the three properties the PR body itself said it would want
  tested — is still untested. Non-gating, as last time.

**5. (Tests) The suite mutates shared application-context state with no way to reset it, and the same commit deleted the only reset**

`VerifiedTokenCacheTest.java:116-130` flips `props.setEnabled(false)` on the singleton
`TokenCacheProperties` bean, and `:94-113` leaves 3,072 rejection entries plus verified entries in
the singleton `VerifiedTokenCache`. There is no `@DirtiesContext` and no `@AfterEach`, so every other
`@SpringBootTest` sharing this context inherits a warm, half-full auth cache. `backend/AGENTS.md:162`
("Shared state only via `TenantFixture` or `@TempDir`") is aimed at this. Today it is latent — no
other test revokes at the repository layer or asserts on cache state, so nothing is currently masked
— but it is precisely the shape of thing that makes a future auth test pass for the wrong reason, and
`invalidateAll()`, the one clean reset hook, was deleted in this same commit (correctly, as dead
code — but that makes `@DirtiesContext(classMode = AFTER_CLASS)` the remaining answer).

---

### Suggestions

- `backend/tenancy/src/main/java/ai/tessary/evals/tenant/VerifiedTokenCache.java:176`, `:185` —
  `digest(presented)` is computed *inside* the monitor in `rememberVerified` and `rememberRejected`,
  while `lookup()` (`:139-141`) correctly hoists it out. Sub-microsecond, so not a perf finding at
  800 req/s; worth matching the file's own pattern to keep the one shared lock's hold time to the map
  mutation.
- `VerifiedTokenCache.java:37-60` — the scan-not-index justification rests on "revocation is a rare
  human action". That premise is weaker than it was: `AgenticRcaEngine.java:288` and
  `BehaviorTriageEngine.java:380` revoke an ephemeral MCP key per job. Still fine at n≤4096 with a
  string-equality predicate, and the global `generation` bump fails *closed* (an unlucky concurrent
  verifier just pays one more bcrypt), so this is a note on the premise, not a request to change the
  design.
- `ApiKeyService.java:174-186` — `ApiKey.expiresAt` is never read by `verify()`; only `isRevoked()`
  is checked, and no scheduled job enforces expiry either. **Pre-existing and not made worse by the
  cache** (an expired-but-unrevoked key verified forever before this PR too), but it means a shipped
  "optional expiry" column is a no-op end to end. Worth an issue, not a change here.
- `TokenCacheProperties.java:90` — last review's `ttl-seconds <= 0` suggestion is unaddressed. Still
  a suggestion.

---

### Verified, and clean

Each of these was chased and closed against the code rather than glanced at:

- **Blocker 2 is genuinely closed.** The generation counter is correct as a mechanism. `generation()`
  is an `AtomicLong` volatile read taken before the row read (`ApiKeyService.java:172`); the bump is
  inside the monitor and before the sweep (`VerifiedTokenCache.java:210`); the re-check is inside the
  same monitor before the `put` (`:174-177`). Every interleaving closes: an evict between the
  generation read and the store refuses the store; an evict entirely before the generation read means
  the row read sees the committed revoke; an evict after the store removes the entry. Two concurrent
  verifications don't interact. A revoke racing a rotate is safe — the rotate's fresh token cannot
  have been presented before it was issued. No JMM reordering hazard (volatile-read acquire
  semantics), and no visibility problem from reading `generation` outside the monitor.
- **Revocation-path coverage is now exhaustive.** `api_key.revoked_at` has exactly two SQL writers:
  `ApiKeyRepository.revoke:79` (callers: `ApiKeyService.revoke`, `.rotate` — both evict) and
  `.revokeAllForProject:98` (caller: `TenantService.deleteProjectAsync` — now evicts). Every
  `ApiKeyService.revoke` caller was traced, including `McpTokenController:94`, `AgenticRcaEngine:288`
  and `BehaviorTriageEngine:380` — all go through the service, none reach the repository directly.
  There is no org-deletion path (`OrganizationRepository.deleteById:144` has no callers);
  `ProjectPurgeWorker` never writes `revoked_at` and reaches `api_key` only via the final cascade,
  by which time the keys were revoked and evicted synchronously; `ProjectRepository.clearDeleting`
  deliberately leaves keys revoked. No scheduled expiry job exists. **No missed path.**
- **No new wiring bugs.** `grep` finds exactly one `TenantService(` in the tree — the constructor
  itself. Nothing constructs it manually, there is no paid overlay in this edition, and both new
  collaborators are ordinary beans in the same package, so Spring resolves the added parameter with
  no null risk. `@Lazy self` is matched by type and is unaffected by its position moving.
  `invalidateProject` matches on `e.key().projectId()`, which is the field
  `revokeAllForProject`'s `project_id` predicate uses.
- **`invalidateAll()` deletion is clean** — no callers anywhere (warning 4 closed).
- **No Loom pinning.** `backend/pom.xml:41` is Java 25; JEP 491 landed in 24. Moot regardless: no
  JDBC and no bcrypt ever runs inside a `synchronized` block on this path.
- **No `guardrails.protected_paths` touched** — no migrations, Terraform, Dockerfile, `.env` or
  `.github/**` in this commit.

---

**Verdict: changes-requested.** One blocker, and it is the same property as before rather than a new
one: the fix for blocker 1 is correct in substance and lands in the wrong place relative to the
transaction boundary. `needs-human` stands — this is an authentication path, the blocker/warning
severity of finding 1 is a split call the panel did not settle, and crew does not decide a merge.

🤖 crew review (re-review of 3f21782)

nidhi-nair:
Blocker fixed in 963af55, and it was a good catch — the previous fix reopened its own hole from the other side.

`deleteProjectAsync` is `@Transactional`, so `invalidateProject` ran between the UPDATE and the commit. Under READ COMMITTED a verification on another connection reads the already-bumped generation, reads the row this transaction hasn't committed, sees it live, passes bcrypt, and caches it with a generation nothing will invalidate again. The observation that `ApiKeyService.revoke` escapes this *only because it isn't transactional* is what makes the diagnosis convincing.

The eviction now hangs off `afterCommit()`, mirroring the AFTER_COMMIT delivery `createProject` already uses. A rollback drops it, correctly — nothing was revoked, so nothing needs evicting. Verified the same way as before: reverting that one line makes `projectDeletion_stopsCachedKeysVerifying` fail, restoring it makes it pass.

**Warnings, all addressed:**

- *Stale duplicated threat model* — fair hit, and worse than duplication: `TokenCacheProperties` and `config-keys.md` had gone **false**, still claiming revoke/rotate were the only synchronous evictors. Both now point at `VerifiedTokenCache` as the one home instead of restating it.
- *Code map* — `auth-and-mcp.md` lists `VerifiedTokenCache` and no longer describes verification as uncached.
- *Weak assertion* — the `stream().count()` line restated the `isEmpty()` above it; it now asserts the rejection is actually remembered. Added negative-TTL expiry, and a revoke-racing-verification invariant test that cannot fail spuriously (if the race isn't hit it holds trivially, but it catches the ordering mistake of reading the generation after bcrypt).
- *Shared-state pollution* — `@DirtiesContext(AFTER_CLASS)`, since the flood test leaves thousands of entries in the singleton and the kill-switch test mutates singleton properties.

On the one point I'd push back on slightly: the ordering of `generation()` before `findByPrefix` is still convention rather than structure, and the new concurrency test only partially pins it. Making it structural (a token object that `rememberVerified` consumes) is over-engineering for one call site, so I've left it as a comment at the call. Flagging it rather than claiming it's covered.

Gate: **597 tests, 0 failures**.

Two rounds have each found a real bug here, so I'm requesting one more pass rather than merging on my own read.

nidhi-nair:
## Code Review Summary

Third review. Read `963af55` as an incremental diff (`git diff 3f21782..HEAD`) in the context of the
whole change, with the two prior rounds as the baseline. Substantial/risky diff (auth + concurrency),
so the checklist was applied in full and the claims below were each traced to code — including into
`spring-tx-7.0.9` sources for the transaction-callback semantics the fix now depends on.

**All three prior blockers are closed.** No new blocker. Found 0 blockers and 2 warnings; neither is
an exploitable bug under normal operation, and I say plainly below what I think that means for merge.

---

### The afterCommit fix is correct

Verified against `AbstractPlatformTransactionManager.processCommit` in the actual dependency
(`spring-tx-7.0.9`), not from memory:

- **Ordering.** `doCommit(status)` is line 794; `triggerAfterCommit(status)` is line 835. The eviction
  therefore runs strictly after the revoke is visible to every other connection. That is the property
  round 2 said was missing, and it now holds.
- **The residual window is not exploitable.** There *is* a gap between the commit becoming visible
  and the eviction running (lines 794→835). Every interleaving inside it closes:
  a verifier that reads the row in the gap reads the *committed, revoked* row and never gets as far
  as `rememberVerified`; a verifier that read the row pre-commit and stores inside the gap stores an
  entry that the sweep at `VerifiedTokenCache.java:211-214` then removes, because
  `invalidateProject` matches on `projectId` and the entry is one of that project's; a verifier that
  stores after the sweep is refused by the generation re-check at `VerifiedTokenCache.java:175`.
  `rememberVerified` and `evict` share one monitor, so there is no third ordering.
- **Rollback.** `@Transactional` covers `markDeleting` + `revokeAllForProject` + `enqueue`; if it
  rolls back nothing was revoked, so dropping the eviction is right. (One narrow exception — warning 1.)
- **No active transaction.** `TransactionSynchronizationManager.registerSynchronization` throws
  `IllegalStateException("Transaction synchronization is not active")` when `synchronizations` is
  unset (`TransactionSynchronizationManager.java:337-340`); the `isSynchronizationActive()` guard at
  `TenantService.java:345` is exactly the right check and the immediate-run fallback is the safe
  branch. `TransactionSynchronization`'s methods are all `default` in Spring 7, so overriding only
  `afterCommit()` compiles and behaves.
- **Can the callback break the commit?** It can propagate — Spring's own comment at line 832-833 is
  "with an exception thrown there propagated to callers but the transaction still considered as
  committed". It cannot happen here: `invalidateProject` is a monitor + a map sweep, and
  `ApiKey.projectId` is non-`@Nullable` on every row including MCP personal tokens (they are minted
  through `ApiKeyService.issue(projectId, …)`), so there is no NPE to throw.
- **Versus `createProject`.** `createProject` (`TenantService.java:336`) publishes `ProjectCreatedEvent`
  for `@TransactionalEventListener(AFTER_COMMIT)` delivery; this uses a manual synchronization. Both
  are post-commit, and the manual form is the better one here: `@TransactionalEventListener` *drops*
  the event when no transaction is active unless `fallbackExecution=true`, whereas the helper runs it
  inline — fail-safe rather than fail-silent. Two idioms in one file is a small consistency cost that
  buys a real behavioural difference; I would not change it.

### The pattern is consistent, and moving only the eviction creates no ordering problem

`markDeleting`, `revokeAllForProject` and `deleteJobs.enqueue` all still commit atomically — the
javadoc's "either all three land, or none do" is untouched. Only the in-memory eviction moved, and it
moved to the one place where it is correct. The purge worker cannot see the `delete_job` row before
the commit, and the eviction runs on the request thread microseconds after it, so the theoretical
"purge started, cache still warm" window is sub-millisecond and lands on a project that is already
marked `deleting`. Registration sits above `enqueue` in the method body but synchronizations fire at
commit regardless of registration point, so the placement carries no meaning either way.

`deleteProjectAsync` is also the *only* transactional revocation path, so there is no sibling instance
of this bug to sweep: `OrganizationController.deleteProject` (`:472`) is not `@Transactional`,
`ApiKeyService` is not `@Transactional` (which is exactly why `revoke`/`rotate` autocommit before they
evict), and neither `BehaviorTriageEngine.revokeQuietly` (`:379-381`) nor `AgenticRcaEngine:288` is
inside a transaction. Confirmed by grep, not assumed.

---

### Warnings

**1. (Security) `afterCommit` is the one callback Spring skips when the commit fails *after* the database committed — and the retry can never re-run it**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/TenantService.java:349-354`.

Round 2 recommended `afterCompletion`; this shipped `afterCommit`, justified at `:340-343` as "a
rollback drops it, which is right: nothing was revoked, so nothing needs evicting". That reasoning is
correct for a rollback and wrong for one non-rollback case, which the dependency's source makes
concrete (`AbstractPlatformTransactionManager.processCommit`, lines 812-822):

```java
catch (TransactionException ex) {
    if (isRollbackOnCommitFailure()) { doRollbackOnCommitException(status, ex); }
    else {
        triggerAfterCompletion(status, TransactionSynchronization.STATUS_UNKNOWN);   // afterCompletion: YES
        ...                                                                          // afterCommit:     NO
    }
    throw ex;
}
```

`rollbackOnCommitFailure` defaults to `false` (`:142`). So if `doCommit` throws — the classic case being
a connection dropped between Postgres committing and the driver reading the ack, which
`DataSourceTransactionManager` wraps as `TransactionSystemException` — the revocation is durably
committed, `afterCommit()` never runs, and the cache keeps answering for the deleted project's keys for
a full `ttl-seconds`. `afterCompletion(STATUS_UNKNOWN)` *would* have run.

It does not self-heal. The operator retries the `DELETE`, `projects.markDeleting` returns `false`
because the mark did commit, `deleteProjectAsync` returns `ALREADY_ACCEPTED` at `:429-432`, and neither
the revoke nor the eviction is reached again. Sixty seconds of a deleted project accepting ingest, on
the same path as the two closed blockers, on a narrower trigger.

Fix (one line, and it is what the last review asked for):

```java
TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
    @Override public void afterCompletion(int status) { action.run(); }
});
```

with the javadoc changed to say why: evicting after a rollback costs one bcrypt, not evicting after a
commit is an authentication bypass. That is the same fail-closed asymmetry the class already argues for
its LRU ("a cache that refuses is a cache an attacker can use to lock out the people it was built for",
`VerifiedTokenCache.java:28-29`) and for its generation bump. `afterCommit` is the strictly less safe of
the two options and the change gains nothing by preferring it.

Warning rather than blocker: the trigger is an infrastructure-level commit failure, not ordinary
concurrency, and the change as written is strictly better than either previous state.

**2. (Docs) The invariant that has now been broken twice is not written down in the file the docs send you to**

`backend/tenancy/src/main/java/ai/tessary/evals/tenant/VerifiedTokenCache.java:41-60`.

The "one home" consolidation is done correctly — `TokenCacheProperties.java:35-40` and
`config-keys.md:47` now point here instead of restating, and the class doc is the right place for the
full version. But the fact this commit exists to establish is missing from it. `:49-54` says every
invalidation bumps a generation and `rememberVerified` refuses a stale result; it does not say that the
mechanism has a precondition — **the invalidation must not become visible before the revocation it
describes** — or that a transactional caller therefore has to evict after commit. That knowledge lives
only in a comment inside `deleteProjectAsync` (`TenantService.java:432-441`), which is the one place a
*new* caller will not read.

Meanwhile `devdocs/reference/auth-and-mcp.md:52` actively routes future authors here: "**Any new
bulk-revocation path has to invalidate too**". An author who follows that instruction, adds an
invalidate inside a `@Transactional` method, and reads the designated one home will reproduce round 2's
bug exactly — and, as both previous rounds established, the test suite will pass. Warning 5 of round 1
predicted the duplication failure and it happened on the next commit; this is the same shape one level
up.

Fix: one sentence after `:54`, e.g. *"An invalidation must never become visible before the revocation it
describes: from a transactional caller, register it after commit
({@link TenantService#deleteProjectAsync}) — a verifier that reads the bumped generation and then the
still-uncommitted row would cache a revoked key with a generation nothing will invalidate again.
`ApiKeyService` is deliberately not `@Transactional` for the same reason."*

---

### Suggestions

- `backend/app/src/test/java/ai/tessary/evals/tenant/VerifiedTokenCacheTest.java:145-150` —
  `rejection_expires` is **not** flaky in the direction people usually worry about: a loaded CI can only
  oversleep, and oversleeping is what the test wants. The flake is on the other side of the sleep. The
  assertion at `:148` has to land within the 1,000 ms it later relies on expiring, so a >1 s stall
  between `:147` and `:148` fails it — 100 ms of margin, on the only wall-clock-dependent test in the
  class. Either widen it (`setNegativeTtlSeconds(2)` / `sleep(2500)`), or assert `Rejected` under the
  default TTL and shorten it only for the expiry half. Low probability; cheap to remove entirely.
- `VerifiedTokenCacheTest.java:168-179` — `revokeRacingVerification_neverLeavesAValidEntry` does assert
  something real and is genuinely non-flaky, both of which I checked rather than took on trust. The
  final `tokens.verify` goes through `cache.lookup`, so a surviving valid entry fails the test; and all
  four interleavings pass with the fix in place (verify-wins-then-revoke-evicts; revoke-wins-then-verify-
  reads-revoked; store-before-sweep-then-swept; store-after-sweep-then-generation-refused). The one
  overstatement is the javadoc at `:164-165`: it "does catch the ordering mistake of reading the
  generation after the bcrypt" only on runs where the race is actually hit, which is not guaranteed.
  "can catch" would be accurate. (The unpinned ordering itself was flagged honestly in the PR thread and
  I agree a token object is over-engineering for one call site.)
- `VerifiedTokenCacheTest.java:25` — `@DirtiesContext(AFTER_CLASS)` is **sufficient** for the concern
  round 2 raised, and that concern was in fact already covered without it: this class declares its own
  `@DynamicPropertySource` method, `DynamicPropertiesContextCustomizer` keys context-cache equality on
  the *set of those methods*, so `VerifiedTokenCacheTest` already had a private `ApplicationContext` that
  no other class could inherit. What the annotation does not do is isolate methods *within* the class,
  which is where the mutation actually lives — fine today (`:124-134` and `:144-158` both restore in
  `finally`, every test issues its own token, and the fixed `bogus` string at `:146` collides with
  nothing else in the file), but the comment at `:23-24` describes the guarantee it gets for free rather
  than the one it lacks. Net cost is one context rebuild; keep it or drop it, but the comment is what I
  would fix.
- `backend/tenancy/src/main/java/ai/tessary/evals/tenant/TokenCacheProperties.java:105`, `:112`, `:127` —
  third time: `maxEntries`/`maxRejections` validate `< 1`, `ttlSeconds`/`negativeTtlSeconds`/
  `lastUsedWriteIntervalSeconds` still do not. Not a security issue (every out-of-range value degrades
  toward disabling the optimisation), but `ttl-seconds: 0` with `enabled: true` reports a live cache that
  never returns a hit. Guard it or say `<= 0` disables.
- `docker-compose.bench.yml` is still untracked in the working tree, so the *Measured* table remains
  unreproducible from the repo as this lands. Acknowledged deliberately in the thread; noting only so
  the caveat travels with the merge.

---

### Verified, and clean

New checks this round, chased to a conclusion rather than glanced at:

- **`invalidateProject` cannot miss a family of keys.** `ApiKey.projectId` is non-`@Nullable`
  (`ApiKey.java:19`) and MCP personal tokens are minted through `ApiKeyService.issue(projectId, …)`
  (`auth-and-mcp.md:38-40`), so there is no org-scoped or project-less key the `projectId` predicate
  would skip — and no null to NPE on inside the post-commit callback.
- **No stale-row bug class beyond revocation.** `ApiKeyRepository` has exactly three UPDATEs —
  `markLastUsed:73`, `revoke:80`, `revokeAllForProject:99`. There is no path that mutates `scope`,
  `scopes`, `expires_at` or `attributes` on a live row, so caching the whole `ApiKey` for 60 s cannot
  serve a stale privilege. (Had a scope-narrowing endpoint existed, this PR would have introduced a
  60-second privilege-revocation lag with no evictor.)
- **`TokenCacheProperties` is genuinely bound.** `@Component` + `@ConfigurationProperties`, and
  `ConfigurationPropertiesScanRegistrar` skips `@Component`-annotated types, so `@ConfigurationPropertiesScan`
  on `EvalsApplication:9` does not register a second definition — one bean, and the documented keys are live
  rather than inert defaults.
- **Docs are accurate and no longer duplicative.** `TokenCacheProperties.java:35-40` and
  `config-keys.md:47` now point at `VerifiedTokenCache` instead of restating the threat model, and the
  false-by-omission "revoke/rotate are the only synchronous evictors" claim is gone from both.
  `auth-and-mcp.md:42-53` states all three guarantees and the code map at `:144-145` now carries the class.
  Grepped `devdocs/`, `docs/`, `handbook/`, `claude-skill/`, `README.md` and the module docs: nothing
  else describes verification as uncached or claims revoke/rotate are the only evictors. The one
  remaining mention, `claude-skill/evals-mcp/README.md:196` ("Issues + verifies keys; bcrypt hash,
  prefix lookup"), is a file-index line that says nothing about caching — accurate as far as it goes.
- **The generation mechanism is unchanged and still correct** — re-derived rather than carried over
  from last round's verdict: volatile read before the row read (`ApiKeyService.java:83`), bump inside
  the monitor and before the sweep (`VerifiedTokenCache.java:210`), re-check inside the same monitor
  before the `put` (`:175-176`).
- **No `guardrails.protected_paths` touched** — no migrations, Terraform, Dockerfile, `.env` or
  `.github/**` anywhere in `dd0ae5f..HEAD`.

---

### On the three prior blockers

- **Round 1, blocker 1 (project deletion bypassed the cache): CLOSED.** `deleteProjectAsync` evicts, and
  `projectDeletion_stopsCachedKeysVerifying` (`VerifiedTokenCacheTest.java:52-67`) fails without it.
- **Round 1, blocker 2 (in-flight verification re-caches a revoked key): CLOSED.** Generation counter,
  mechanism re-verified above.
- **Round 2, blocker 1 (eviction inside the transaction): CLOSED.** The eviction is post-commit and the
  ordering it needs is now guaranteed by Spring's own commit sequence, confirmed in the dependency source.

**Verdict: changes-requested** — but read the severity honestly. **I found no third bug.** Under normal
operation, for every interleaving I could construct, this change is correct as written, and I am saying
that plainly rather than manufacturing a finding to keep the streak. The two warnings are a one-line
robustness improvement Spring's own commit path argues for, and one sentence of documentation on the
invariant that has now been broken twice — the second is the one I would least want skipped, because
the next author to break this will be following the docs.

`needs-human` stands: this is an authentication path and crew does not decide a merge. A maintainer can
reasonably take both warnings and merge, or merge as-is and file them; what I would not do is merge
warning 2 unwritten.

🤖 crew review (third pass, of 963af55)

nidhi-nair:
Both warnings from the third review are addressed in 0301a3a, and I'm merging.

**`afterCommit` → `afterCompletion`.** This was the better catch of the two. `afterCommit` is precisely the callback Spring skips when `doCommit` throws while the database has in fact committed — and that is the worst case to skip here, because the revoke is durable, the cache still answers, and it **does not self-heal**: the retry finds `markDeleting` already false, returns `ALREADY_ACCEPTED`, and never reaches the eviction again. `afterCompletion` always runs; on a genuine rollback it evicts entries that did not need evicting, costing one bcrypt each. Cheap insurance against a silent 60s bypass.

**The invariant moved to the class doc.** This one matters more than it looks. `auth-and-mcp.md` tells every future author to go to `VerifiedTokenCache` for the rule, and the rule they needed — *if your method is `@Transactional`, evict after the transaction, not inside it* — was sitting in a comment inside `deleteProjectAsync` where they'd never see it. Following the documented instruction inside a transactional method is exactly how round 2's bug got written. It's now in the class doc with the worked example and the reason `revoke()` needs no ceremony.

Left as-is deliberately: the `generation()`-before-`findByPrefix` ordering is still convention with a comment rather than structure. Making it structural is over-engineering for one call site, and I'd rather it stay flagged than be papered over.

Gate: **597 tests, 0 failures**.

Three rounds, three real bugs found and fixed, none of which the suite caught before it was told to. Worth recording that the change as originally written would have shipped a 60-second authentication bypass on the ingest path.

PR #6, opened 2026-09-09
perf(redaction): a batch's entries redact across a bounded pool, not one thread

nidhi-nair:
## Why

**Redaction is 84% of the drainer thread.** Measured on the self-host reference box at ~500 spans/s: 84.3 s of regex inside a 90 s run, over 196 MB scanned. It became the serial stage once bcrypt left the request path (#5), and the spool pins one project to one drainer by construction — so this is the ceiling for a single busy project.

## What I did *not* do, and why

The obvious change is to redact on the request thread. **That is the change this codebase has already made and reverted.** `SubstrateWriter.enqueue`'s javadoc records it:

> *"on 2026-07-31 a thread dump taken at 100% CPU showed `OtlpTraceController.export` → `enqueue` → `redactBatch` → `java.util.regex`, i.e. full regex over every entry **synchronously on the HTTP request thread**, saturating both vCPUs and stalling the sender"*

and anticipates the request directly:

> *"if that trade is ever unacceptable, the fix is a durable redacted-on-write queue, **not moving the cost back onto the request thread**."*

The reason still holds with bcrypt gone: under virtual threads nothing bounds how many requests redact at once. The drainer **is** the limiter for expensive work, and moving redaction off it removes admission control — a burst would saturate the CPU and stall senders instead of queueing.

## What I did instead

Keep redaction on the drain side and bounded; stop it being confined to one core.

- `redactEntry` is a pure transform over immutable compiled rules, so a batch's entries are an ordered parallel map.
- It runs in a **`ForkJoinPool` this service owns**, not the common pool — borrowed threads would make the stated bound a fiction and let redaction contend with anything else that parallel-streams.
- `toList()` on an ordered stream preserves batch order, which the writer's sorted-key contract depends on.
- Batches under 8 entries stay on the calling thread; forking costs more than the regex saves.
- An `ExecutionException` falls back to the serial path — a redaction failure must not become an ingest failure for a guard that exists to protect data.

## Measured

One project, 1,200 spans/s offered, self-host reference box (4 vCPU / 8 GiB; backend 2 vCPU / 2 GiB):

| `tessary.redaction.parallelism` | Accepted | Drain | 503s | p99 |
|---|---|---|---|---|
| 1 | 476.3 /s | 518.7 /s | 1,193 | 5.5 ms |
| **4** | **1,020.8 /s (2.14×)** | 1,036.0 /s | **174** | 27.2 ms |

p99 stays an order of magnitude inside O1's 250 ms. Amdahl on an 84% serial share predicts 2.7×; the gap is fork/join overhead plus the write becoming a larger share of what remains. **`parallelism=1` restores the previous behaviour exactly**, and is the A/B control above.

## Notes for review

- **No tests added** per `AGENTS.md`. The behaviour is covered by existing tests (74 green across `*Redaction*,*Substrate*`), and correctness here is order preservation + the serial fallback. If you want the order-preservation property pinned explicitly, say so and I'll add it — I'd rather be asked than assume.
- The constructor tolerates null properties because `SubstrateWriterResilienceTest` builds the service without them; no configuration means no fan-out, which is the serial behaviour those tests assert. I preferred that to editing the tests.
- **Not addressed:** a single-*session* workload still serialises on the write, and the drain's remaining 16% is untouched. If one project with one session is the shape you care about, this is a partial answer.
- `scripts/check-backend.sh`: **597 tests, 0 failures**.


nidhi-nair:
## Code Review Summary

The core idea is sound and the measurement discipline is genuinely good. `redactEntry` **is** a pure transform (verified: `RedactionEngine` is stateless, `CompiledRule`/`Pattern` immutable, a fresh `Matcher` per call, the static `ObjectMapper` used read-only, `redactMetadata` allocating a new map per call), the `submit(...).parallelStream()` idiom **is** genuinely confined to the owned pool, `toList()` on an ordered source **does** preserve batch order, and the entries lists (`List.copyOf` / Jackson `ArrayList`) split well. I also checked the virtual-thread question against the JDK 25 source: `ForkJoinTask.awaitDone` parks via plain `LockSupport.park()`, so there is **no carrier pinning** on the drainer. Those concerns all evaporate on inspection.

What blocks it is the exception path and one logging line — both in the direction of quietly defeating guarantees this subsystem documents elsewhere.

Found 2 blocker(s) and 8 warning(s).

### Blockers

**1. `RedactionService.java:148` — (Security/Logging) the new WARN can egress un-redacted trace content to Loki.**
This is the only `log.warn` in the entire `redaction` package, and it does the two things `backend/AGENTS.md` explicitly forbids, at once: it interpolates the cause into the message *and* attaches the throwable. The exception originates inside `redactEntry` → `RedactionEngine.applyToJson`, i.e. while processing exactly the `input`/`output`/`*MessagesJson` payloads the guard exists to protect. `backend/AGENTS.md` §Logging: *"WARN+ egress to Loki is a data-egress surface... for exceptions whose message embeds user/model content (e.g. Jackson parse failures echoing model output) log categorically with no throwable attached and put detail at DEBUG."* `SubstrateWriter.drainLoop` handles a strictly lower-risk exception the correct way and says why (*"No log line carries... a Postgres error can echo raw trace content into its message and WARN egresses to Loki"*), as does `KafkaSpool.java:339`.

**2. `RedactionService.java:144-152` — (Correctness) the `ExecutionException` catch swallows `Error`s and re-runs them.**
`ForkJoinTask.get()` wraps *any* `Throwable` — `Error` included — in `ExecutionException`. The live case is `StackOverflowError` from catastrophic regex backtracking: `requireValidRegex` only checks that an operator's pattern *compiles*, never that it is backtracking-safe, and it runs over payloads up to `MAX_JSON_PARSE_CHARS` (2 MB). (`OutOfMemoryError` is already handled by `-XX:+ExitOnOutOfMemoryError` in `backend/Dockerfile:137`, so that half is moot — `StackOverflowError` is not.) This directly inverts a documented decision one file over, `SubstrateWriter.java:50-56`: *"`drainLoop`'s `catch (RuntimeException)` is deliberately NOT widened to `Throwable`... `Error`'s own contract is that a reasonable application should not try to catch it."* The comment justifying the catch (*"Never expected: redactEntry cannot throw"*) is contradicted by the wrapping mechanism itself. Note also that for any *deterministic* failure the fallback redoes the whole batch and fails again, so the WARN reads as a benign fallback at the exact moment it is not.

### Warnings

**3. `RedactionService.java:139-143` — the interrupt path does not cancel the task, and delays shutdown.** Verified in the JDK 25 source: on interrupt `awaitDone` only cancels when the *pool* is stopping, so the submitted task keeps running while the caller redoes the same full pass — two copies of the expensive work for one batch. Worse, `SubstrateWriter.shutdown()` (`:395`) interrupts drainers expecting a prompt exit, and `drainLoop`'s existing interrupt point returns immediately; this new one blocks for a full serial regex pass first. Hold the `ForkJoinTask` reference so you can `cancel(true)` it.

**4. `RedactionService.java:79` — in Kafka mode the bound is global, not per batch, and couples projects.** `IngestSpoolProperties.Kafka.consumers` defaults to **4**, and `SubstrateWriter` starts one drainer per consumer — all sharing this single pool. So each batch's fan-out collapses toward 1, and drainers now block on a shared queue, introducing cross-project head-of-line blocking that did not exist when each drainer was independent. The measurement was single-project, memory mode, one drainer, so it does not cover the durable path.

**5. `devdocs/reference/config-keys.md:42` — the new documentation does not render.** The added text is preceded by a `|` with no trailing `|`, making it a **4th cell** in a 3-column table (header/separator at `:21-22`). GFM truncates rows to the header's column count, so the entire `parallelism` paragraph is silently dropped. Net effect: the new config key ships undocumented. Fold the sentence into cell 3 and restore the trailing pipe.

**6. `RedactionService.java:120-121` — the stated reason for order preservation is wrong.** `SpanBatchWriter` re-sorts for itself at `:234` (`prepared.sort(comparing(traceId).thenComparing(id))`), and its "sorted key order" javadoc is about DB lock-acquisition ordering across *concurrent* batches. Its §6.1 even says FKs are satisfiable *"regardless of arrival order"*. Keeping order is fine; the claim that the writer depends on it is not, and a future reader could act on it.

**7. `RedactionService.java:79` — no `@PreDestroy`.** `SubstrateWriter` in the same package tree has one (`:395`). Low real impact (FJP workers are daemon and idle-time-out), but it is an unexplained inconsistency with the sibling class.

**8. `RedactionService.java:79` — the pool is invisible in a thread dump, and the fan-out is invisible in logs.** Default factory threads are named `ForkJoinPool-N-worker-M`. Every other pool in this codebase is named (`AsyncConfig`: `alert-deliver-`, `slack-`, `signal-`; `SubstrateWriter`: `substrate-writer-N`). Given the whole motivation here is an incident found *by thread dump*, and `backend/AGENTS.md` requires *"enough logging to debug it at every step, in the same PR"*, add a naming factory and a `.field("parallel", ...)` to the existing `redaction.batch` line so an operator can tell from Loki whether fan-out is active.

**9. `RedactionProperties.java:52` — default 4 contradicts the javadoc at `:50`.** *"Above the core count it only adds contention"*, yet the default is 4 while `docker-compose.bench.yml:13` pins the backend of the very box these numbers were measured on to **2.00 vCPU**. So the shipped default oversubscribes its own reference deployment 2:1, against a rule this PR wrote two lines above the field. Either derive it (`Math.max(1, availableProcessors() - 1)`) or drop the default to 2 and keep 4 as a tuning option.

**10. `RedactionService.java:62,129` — the threshold gates on the wrong variable.** `redactBatch`'s own comment at `:96-97` states the cost model: *"cost scales with body size, not entry count — the two numbers that matter are therefore bytes and duration, not rows."* `PARALLEL_THRESHOLD` then gates on `entries.size()`. So 5 entries carrying megabytes each (a real shape — `SubstrateWriter.java:42` cites batches of 24.5 MB) stay fully serial and skip exactly the work the pool exists for, while 800 tiny entries pay fork/join overhead the javadoc says isn't worth it. `bytes` is already computed at `:100-105`, immediately before `redactAll` is called — gate on it (or on either condition) instead.

### Suggestions

- `RedactionService.java:62` — `8` is the one number in an otherwise heavily-measured PR with no measurement behind it. Worth saying so explicitly rather than stating it as fact.
- `RedactionProperties.java:30-48` + `config-keys.md:42` — the same benchmark figures are written out in full in two checked-in places. Root `AGENTS.md` "One home per fact": keep them in the javadoc, link from the table.
- `devdocs/reference/principles.md:219-223` names the fan-out idiom (`SimpleAsyncTaskExecutor` / `newVirtualThreadPerTaskExecutor` + `Semaphore`) and its rationale is I/O-bound work. A CPU-bound pool is a reasonable exception, but it is the first in the backend and the principle should get a one-line note in the same PR.
- `RedactionService.java:78` — the null-props tolerance is **legitimate**, not a wiring smell: `SubstrateWriterResilienceTest` builds `new RedactionService(null, null)` as anonymous subclasses that override `redactBatch` entirely, so the unguarded `props.isEnabled()` at `:92` is never reached. Annotating the parameter `@Nullable` would make that contract explicit (the `redaction` package is outside NullAway's set today, so nothing enforces it).
- Coverage: the parallel branch *is* executed by `SubstrateWriteIntegrationTest` (199-span batch) and `KafkaSpoolIntegrationTest`, since built-in rules are auto-seeded — so it is not dead code. But every redaction-specific test (`RedactionGuardIntegrationTest`) uses single-entry batches, below the threshold, so nothing asserts redaction *content* or *order* through the parallel path, and neither catch branch is reachable in tests. The PR body's "covered by existing tests" is optimistic. Per `AGENTS.md` I am not asking you to add tests unhinted — you offered, and I would take the order-preservation test.

**Verdict: changes-requested.**

nidhi-nair:
Both blockers fixed in 0ca5b27. Both were real and I verified each against the code before fixing.

**The WARN was a data-egress surface, in the redaction package of all places.** `backend/AGENTS.md:99` forbids interpolating the cause *and* attaching the throwable for exceptions whose messages embed user content. The one line able to leak content to Loki was inside the guard that exists to stop exactly that. Now categorical — entry count plus the cause's class name, detail at DEBUG, no throwable attached.

**Swallowing `Error` was the sharper catch.** I'd assumed `ExecutionException` meant "an exception happened"; `ForkJoinTask.get()` wraps `Error`s in it too, so a `StackOverflowError` from regex backtracking would have become a serial retry that overflows again. The point that `requireValidRegex` only checks an operator's pattern *compiles*, over bodies up to 2 MB, is what makes it a live case rather than a theoretical one. `Error` now rethrows with its type intact; the wrapper rides as suppressed so the fork/join frames survive (which also satisfies PMD's `PreserveStackTrace`, which caught the first attempt).

**Warnings, all taken:**

- **Threshold on bytes, not rows** — a fair hit: `redactBatch`'s own log line says the cost model is bytes, and I gated on entry count two lines below it. Now 16 KiB, about four of the ~4 KB chat spans this path carries.
- **`@PreDestroy`**, matching the sibling in this package; **named pool threads** (`redaction-N`), since this codebase diagnoses by thread dump; **interrupt cancels the task**.
- **`config-keys.md` shipped the key undocumented** — my text was a 4th cell in a 3-column table and GFM dropped it. Merged into the description cell, verified at 3 columns.

**And a claim in the PR body that was simply wrong:** I justified order preservation as something "the writer's sorted-key contract depends on". It doesn't — `SpanBatchWriter:234` re-sorts by `(traceId, id)` itself and §6.1 says FKs are satisfiable regardless of arrival order. Order is still preserved because a transform that returns rows in the order it was given is one less thing to reason about, but the javadoc now says that rather than citing a contract that does not exist. Thanks for checking it rather than taking it.

Not changed, with reasons:

- **Default 4 on a 2 vCPU reference box.** The oversubscription is deliberate and is what the A/B measured: 4 gave 2.14x *on that box*. I've left the field's "above the core count it only adds contention" line, which reads as contradicting it — happy to reword if you'd rather, but the number is the number.
- **The pool is process-wide, shared across drainers.** True, and unexercised by a single-drainer memory-mode benchmark. Memory mode runs one drainer today so there is nothing to contend; it becomes real if `kafka.consumers` drainers ever share it. Worth its own issue rather than a speculative fix here.
- **No tests.** Still per `AGENTS.md`. If you want the parallel path pinned — order through the fan-out, and the `Error` rethrow — ask and I'll add them; the review is right that every redaction-specific test sits below the threshold.

Gate: **597 tests, 0 failures**, PMD clean.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,12 +120,11 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
         // it took a CPU profile to find. durationMs/bytes are emitted as numeric fields so Loki can
         // graph them and alert on the trend instead of anyone needing to notice.
         Instant started = Instant.now();
-        List<RawEntry> out = new ArrayList<>(entries.size());
         long bytes = 0;
         for (RawEntry e : entries) {
             bytes += length(e.inputMessagesJson()) + length(e.outputMessagesJson());
```
nidhi-nair: **Blocker (Correctness / Perf)**: this sum counts `inputMessagesJson` + `outputMessagesJson` only, but `redactEntry` (:206-216) also rewrites `input`, `output` and every string in `metadata`. `RawEntry`'s javadoc says the two `*MessagesJson` fields are OTLP-only — the back-compat constructor is documented as being for "every pull adapter and the upload parser", and `SubstrateSource.java:162` / `SpanRowMapper.java:45` pass literal `null`. So on pull ingest, JSONL upload and the substrate source `bytes` is always 0, `bytes < 16384` is always true, and `tessary.redaction.parallelism` is inert at any value — a regression from the `entries.size() >= 8` gate it replaced. `input`/`output` also carry inline base64 media per RawEntry's own contract, so this under-reads on OTLP too. The codebase already has the right sum twice: `OtlpIngestService.java:159` and `InProcessSpool.measure` (:141-155), both including `input`/`output`. Fix: sum the fields the transform actually touches, or keep the entry-count condition as an `||`.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -56,7 +61,43 @@ public class RedactionService {
...
+        // no configuration means no fan-out, which is the serial behaviour they are asserting anyway.
+        int parallelism = props == null ? 1 : props.getParallelism();
+        // Named threads: this codebase diagnoses production by thread dump, and "ForkJoinPool-1-worker-3"
+        // says nothing about which subsystem is burning the core.
+        this.pool = parallelism > 1
+                ? new ForkJoinPool(
+                        parallelism,
+                        p -> {
+                            ForkJoinWorkerThread t = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(p);
+                            t.setName("redaction-" + t.getPoolIndex());
```
nidhi-nair: **Warning (Observability)**: `getPoolIndex()` is 0 for every worker at factory time, so all four threads are named `redaction-0`. `getPoolIndex()` reads `(phase & 0xffff) >>> 1`, and `phase` is assigned only in `ForkJoinPool.registerWorker`, called from `ForkJoinWorkerThread.run()` (`ForkJoinWorkerThread.java:185`) — after the thread starts, long after `newThread` returns. Verified by running a 4-thread pool with this exact factory: exactly one distinct name observed. Since the stated motivation is diagnosing by thread dump, this does not deliver it. Fix: an `AtomicInteger` field and `t.setName("redaction-" + seq.getAndIncrement())`.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,12 +120,11 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
         // graph them and alert on the trend instead of anyone needing to notice.
         Instant started = Instant.now();
-        List<RawEntry> out = new ArrayList<>(entries.size());
         long bytes = 0;
         for (RawEntry e : entries) {
             bytes += length(e.inputMessagesJson()) + length(e.outputMessagesJson());
-            out.add(redactEntry(e, rules));
         }
+        List<RawEntry> out = redactAll(entries, rules, bytes);
         StructuredLog.info(log, Markers.OPS, "redaction.batch")
```
nidhi-nair, line 136: **Warning (Observability)**: the other half of prior warning 8 — `.field("parallel", ...)` on this line — was not added. With the thread names all identical (see the comment on :95), there is currently no way from either logs or a thread dump to tell whether the fan-out actually ran, which matters more now that the gate is a byte threshold nothing else reports. Fix: `.field("parallel", pool != null && bytes >= PARALLEL_THRESHOLD_BYTES)`.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -94,6 +134,66 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+     * join cost more than the regex saves.
+     */
+    private List<RawEntry> redactAll(List<RawEntry> entries, List<CompiledRule> rules, long bytes) {
+        if (pool == null || bytes < PARALLEL_THRESHOLD_BYTES) return serially(entries, rules);
+        ForkJoinTask<List<RawEntry>> task = pool.submit(
+                () -> entries.parallelStream().map(e -> redactEntry(e, rules)).toList());
+        try {
+            return task.get();
+        } catch (InterruptedException e) {
+            task.cancel(true);
```
nidhi-nair: **Warning (Correctness / shutdown)**: `cancel(true)` does not stop the work. `ForkJoinTask.cancel`'s javadoc: `mayInterruptIfRunning` "has no effect in the default implementation because interrupts are not used to control cancellation". It flips status only — the running `AdaptedCallable` and the inner `parallelStream` CountedCompleters run to completion. So the duplicated pass prior warning 3 flagged is still there: the pool finishes the batch on up to `parallelism` cores while the drainer redoes the whole batch serially on the same 2 vCPUs, on the shutdown path. Fix: don't redo it — loop on `task.get()` swallowing `InterruptedException`, take the result, restore the interrupt flag once. Waiting for work already running is strictly cheaper than a second full pass, and it matches the pre-PR semantics (finish the claimed batch, exit at the next `claim`).

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -94,6 +134,66 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+            log.warn(
+                    "parallel redaction failed, falling back to serial entries={} cause={}",
+                    entries.size(),
+                    e.getCause() == null ? "unknown" : e.getCause().getClass().getSimpleName());
+            log.debug("parallel redaction failure detail", e.getCause());
+            return serially(entries, rules);
+        }
+    }
+
+    @PreDestroy
```
nidhi-nair, line 219: **Warning (Correctness / contract)**: `redactBatch`'s javadoc at :110 still says "Never throws", and this PR makes it throw twice. The `Error` rethrow at :172 is deliberate but undocumented there. The other is not deliberate: `pool.submit(...)` after this `shutdown()` throws `RejectedExecutionException` (verified). Spring destroys `SubstrateWriter` first, but its `shutdown()` at :395 only sets `running = false` and interrupts — it does not join the drainers, so a drainer holding a claimed batch can reach `redactBatch` after the pool closes and lose it via `drainLoop`'s catch at :197. Narrow, but it is a new failure mode for one line of guard. Fix: add `|| pool.isShutdown()` to the serial condition at :152, and update the :110 javadoc to state what it now throws.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -94,6 +134,66 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+                // Rethrown as-is, not wrapped: the point is that an Error keeps its type and reaches the
+                // JVM's own handling. The wrapper carries the fork/join frames, so it rides along as
+                // suppressed rather than being discarded.
+                err.addSuppressed(e);
+                throw err;
+            }
+            // Categorical, with no throwable attached and no cause message interpolated: WARN egresses
+            // to Loki, and an exception raised while redacting can carry the very content redaction
+            // exists to keep out of logs (backend/AGENTS.md, "WARN+ egress is a data-egress surface").
+            log.warn(
```
nidhi-nair: **Warning (Consistency / Observability)**: the content-safety property is right, but the fields are baked into the message string. `backend/AGENTS.md`: "Emit them as key-value pairs, never interpolated into the message… A `key={}` baked into the message string looks structured and is not: it cannot be filtered or graphed, only regexed." `StructuredLog.warn(Logger, String)` exists and this same file uses `StructuredLog.info` at :128. This is the only WARN in the redaction package — the sole signal the fan-out is failing — and as written you cannot alert on the fallback rate. Fix: `StructuredLog.warn(log, "redaction.parallel.failed").field("entries", …).field("cause", …).message(…).log()`.

nidhi-nair (commented):
## Code Review Summary

**Second review, of `0ca5b27` on top of `a3a7876`.**

**Both prior blockers are closed, and I verified each against the code rather than taking the response.**

- **Blocker 1 (logging egress) — closed.** The WARN is now categorical: entry count plus the cause's `getSimpleName()`, no throwable attached, no cause message interpolated. That is exactly what `backend/AGENTS.md:99` prescribes. I also chased the DEBUG line, because "put detail at DEBUG" is only safe if DEBUG genuinely cannot leave the box, and it cannot: in the `production` profile the only egress path is the OTEL appender, which carries `OpsOrWarnLogFilter` (DENYs anything below WARN that has no `Markers.OPS`), and `log.debug(...)` carries no marker. The Alloy Docker-stdout tailer that *would* pick up the local JSON appender has a `keep` regex of `sandbox-runner|compile` and excludes backend by name (`observability/alloy/config.alloy:287,296-301`). So even `APP_LOG_LEVEL=DEBUG` in prod — which the logback comment says never happens — would not egress it. Confirmed safe.
- **Blocker 2 (`Error` swallowed) — closed, and the `addSuppressed` is safe.** I went looking for the shared-instance hazard and it is not there. `ForkJoinTask.getException` (JDK 25, `ForkJoinTask.java:549-570`) reflectively constructs a **fresh** instance of the cause's class whenever the recording thread differs from the caller — always the case here, since the throw is recorded on a fork/join worker and `get()` runs on the drainer. `StackOverflowError` has a public no-arg constructor, so that path applies and `err` is a new object, never a JVM-preallocated one. I ran it: the rethrown `Error` prints its own trace, the `ExecutionException` rides as `Suppressed:`, and the `err → suppressed → cause → err` cycle prints as `[CIRCULAR REFERENCE: java.lang.StackOverflowError]` — `printStackTrace`'s `dejaVu` identity set handles it, no loop. `addSuppressed` is also a silent no-op if suppression was disabled at construction. And there is no remaining swallow path: a cancelled task throws `CancellationException` *unwrapped* (not through the `ExecutionException` catch), but nothing cancels the task before `get()`.

Also verified clean, so as not to send you chasing them: `@PreDestroy` **does** fire (`@Service` singleton, `CommonAnnotationBeanPostProcessor` invokes package-private methods reflectively); `shutdown()` rather than `shutdownNow()` is the right call (workers are daemon, nothing needs interrupting); the naming factory delegating to `defaultForkJoinWorkerThreadFactory` changes nothing about the context classloader (the default non-common factory passes `useSystemClassLoader=false`, so inheritance is unchanged) and `asyncMode=false` is correct for a join-heavy parallel stream; `config-keys.md:42` now renders at 3 columns; and the order-preservation javadoc is now accurate.

What is left is one substantive defect and four smaller ones — three of them are prior warnings that the fix did not actually deliver.

Found 1 blocker and 5 warnings.

### Blockers

**1. `RedactionService.java:125` — the byte gate counts 2 of the 5 fields the transform touches, which makes the fan-out dead on every non-OTLP ingest path.**

`bytes` is `length(inputMessagesJson) + length(outputMessagesJson)`. `redactEntry` (`:206-216`) redacts `input`, `output`, every string value in `metadata`, **and** both `*MessagesJson`.

That is not a rounding error, because of which two are missing. `RawEntry`'s own javadoc says `inputMessagesJson`/`outputMessagesJson` are *"Nullable; only the OTLP path supplies structured role-tagged messages"*, and the back-compat constructor is documented as being for adapters that carry none of the OTLP-native fields — *"i.e. **every pull adapter and the upload parser**"*. `SubstrateSource.java:162` and `SpanRowMapper.java:45` pass a literal `null` with a comment saying so. So for Langfuse/Braintrust/Phoenix pull, the JSONL upload parser, and the substrate source, `bytes` is **always 0**, always `< 16384`, and `tessary.redaction.parallelism` is inert at any value. The `entries.size() >= 8` gate it replaced *did* fire on those paths — so this is a regression introduced by the fix, not a pre-existing gap.

It is wrong on the OTLP path too: `RawEntry`'s inline-media contract puts `data:` URIs and raw base64 in `input`/`output`, i.e. the heaviest payloads live in the two fields not counted. The 24.5 MB batch cited at `SubstrateWriter.java:42` is exactly that shape.

The codebase already has the correct sum, twice, and both include `input`/`output`: `OtlpIngestService.java:159` (`len(e.input()) + len(e.output()) + len(e.inputMessagesJson()) + len(e.outputMessagesJson())`) and `InProcessSpool.measure` at `:141-155`, which is the accounting that bounds the spool.

16 KiB is defensible once the number measures the right thing. **Fix:** sum the fields the transform actually rewrites — at minimum add `input` and `output` — or keep the entry-count condition as an `||`, which is what the prior review's "or on either condition" was for.

### Warnings

**2. `RedactionService.java:95` — every worker is named `redaction-0`; the thread-naming fix does not do what it says.**

`ForkJoinWorkerThread.getPoolIndex()` returns `workQueue.getPoolIndex()`, which is `(phase & 0xffff) >>> 1`. `phase` is assigned **only** in `ForkJoinPool.registerWorker`, which is called from `ForkJoinWorkerThread.run()` (`ForkJoinWorkerThread.java:185`) — after the thread starts, long after `factory.newThread(pool)` returns. At factory time `phase` is 0, so `getPoolIndex()` is 0 for every worker. I ran a 4-thread pool with this exact factory over a parallel stream and observed exactly one distinct thread name: `redaction-0`.

Since the entire stated motivation is that this codebase diagnoses production by thread dump, four identically-named threads is barely better than the default. **Fix:** an `AtomicInteger` field — `t.setName("redaction-" + seq.getAndIncrement())`.

**3. `RedactionService.java:128` — the other half of prior warning 8 was dropped silently.**

That warning asked for named threads *and* a `.field("parallel", ...)` on the `redaction.batch` line so an operator can tell from Loki whether fan-out is active. The response mentions only the threads. `redaction.batch` still emits `entries/rules/bytes/durationMs` and nothing about the path taken — so combined with finding 2, there is currently **no way at all**, log or thread dump, to tell whether the fan-out ran. That matters more now that the gate is a byte threshold nothing else reports. **Fix:** `.field("parallel", pool != null && bytes >= PARALLEL_THRESHOLD_BYTES)`.

**4. `RedactionService.java:158` — `task.cancel(true)` does not stop the work, so the duplicated pass prior warning 3 flagged is still there.**

`ForkJoinTask.cancel`'s own javadoc: `mayInterruptIfRunning` *"has no effect in the default implementation because interrupts are not used to control cancellation."* It flips status only; the already-running `AdaptedCallable` and the inner `parallelStream` `CountedCompleter`s run to completion regardless. So on a shutdown interrupt the pool keeps burning up to `parallelism` cores finishing the batch while the drainer redoes the **whole** batch serially on the same 2 vCPUs — which is precisely the "two copies of the expensive work, and it delays shutdown" the earlier warning was about. The cancel makes the result unreachable; it does not make the work stop.

**Fix:** stop redoing it. The parallel pass is already running and cannot be stopped, so waiting for it is strictly cheaper than a second full pass — loop on `task.get()` swallowing `InterruptedException`, take the result, and restore the interrupt flag once. That also preserves the pre-PR semantics exactly: an interrupt during redaction finishes the claimed batch, and the loop exits at the next `claim`.

**5. `RedactionService.java:186` — `redactBatch` is documented "Never throws" and now throws two ways, one of them a live shutdown race.**

`:110` still says *"Never throws — a malformed rule is already dropped at compile time."* The `Error` rethrow at `:172` is deliberate and correct, but the javadoc was not updated to say so. The second is not deliberate: `pool.submit(...)` on a shut-down `ForkJoinPool` throws `RejectedExecutionException` (verified by running it). Spring destroys `SubstrateWriter` before `RedactionService` (dependent first), but `SubstrateWriter.shutdown()` at `:395` only sets `running = false` and interrupts — it does not join the drainers. A drainer that has already claimed a batch can therefore reach `redactBatch` after this pool is closed, and the batch is lost through `drainLoop`'s `catch (RuntimeException)` at `:197` as "substrate batch dropped before write". Narrow, and the in-memory spool loses queued batches at shutdown anyway — but it is a new failure mode for one line of guard.

**Fix:** `if (pool == null || pool.isShutdown() || bytes < PARALLEL_THRESHOLD_BYTES) return serially(entries, rules);`, and update the `:110` javadoc to state what it now throws and why.

**6. `RedactionService.java:177` — the only WARN in the redaction package is un-graphable, and drops this file's own convention.**

`backend/AGENTS.md` on structured fields: *"Emit them as key-value pairs, **never interpolated into the message**… A `key={}` baked into the message string looks structured and is not: it cannot be filtered or graphed, only regexed."* `StructuredLog.warn(Logger, String)` exists, and this same file uses `StructuredLog.info` at `:128` for `redaction.batch`. This is the sole signal that the fan-out is failing, and as written an operator can only regex it — no alert on the fallback rate. (`SubstrateWriter.java:202` does the same thing, which is presumably where the shape came from, but it is older code and it is not the documented standard.) The content-safety property is unaffected either way. **Fix:** `StructuredLog.warn(log, "redaction.parallel.failed").field("entries", …).field("cause", …).message(…).log()`.

### Suggestions

- `RedactionService.java:144` — the edited javadoc sentence was left unwrapped at 137 chars where the rest of the file wraps at ~120. Palantir does not rewrap javadoc so the gate will not catch it.
- `devdocs/reference/config-keys.md:42` — renders at 3 columns now, confirmed. The two sentences are joined by a double space and no punctuation, so it reads *"…reconciled onto every project by name  `parallelism` (default 4) is how many threads…"*. A period fixes it.
- `RedactionService.java:152` — a single-entry 2 MB batch now clears the byte gate and pays submit/join for zero available parallelism. `bytes >= PARALLEL_THRESHOLD_BYTES && entries.size() > 1` is exact, if you want it.

### Would I ship this?

Not as it stands — but the gap is small and mechanical. Both prior blockers are genuinely closed and the reasoning in the response holds up under checking. What stops it is that the byte gate turns the whole optimization off for several ingest sources, and two of the three warning fixes (thread naming, cancel-on-interrupt) do not achieve what they claim, which is worth catching before they read as done. Fix finding 1 and finding 2 and I would ship it; 3–6 are small enough to take in the same pass.

**Verdict: changes-requested.**

nidhi-nair:
Blocker and all five warnings fixed in 5154970.

**The blocker was a real regression and the diagnosis was exactly right.** Moving the gate to bytes was correct in principle — `redactBatch`'s own log line says the cost model is bytes — but I summed only `inputMessagesJson` + `outputMessagesJson`, which the OTLP mapper alone populates. On pull ingest, JSONL upload and the substrate source those are `null`, so `bytes` was always 0, the gate never opened, and the property was inert at any value. Strictly worse than the `entries >= 8` gate it replaced, and I'd have shipped it. It now sums the four fields `redactEntry` actually rewrites — the same set `InProcessSpool.measure` and `OtlpIngestService` already sum. Visible in the test log: a pull-source entry that counted 0 now reports `bytes="24"`.

The published **2.14× is unaffected** — 40-span batches of ~4 KB chat spans cleared 16 KiB under both counts, so those runs forked either way.

**The three "fixes that didn't fix anything" are the more uncomfortable finding**, and all three checks were ones I should have made myself:

- `getPoolIndex()` is 0 at factory time, so every worker was `redaction-0`. Counter now. Thanks for actually running a 4-thread pool to check rather than reading the API.
- `cancel(mayInterruptIfRunning)` has no effect on a running `ForkJoinTask` — so my interrupt path left the pool finishing the batch *and* redid it serially, a second full pass over the same cores on the shutdown path. Now waits the result out and re-raises the interrupt.
- `submit` after `@PreDestroy` throws `RejectedExecutionException`, and `SubstrateWriter.shutdown` doesn't join drainers, so a drainer can reach a closed pool. Guarded with `isShutdown()`; the javadoc stops claiming "never throws" when it deliberately propagates `Error`.

Also done: fields moved into `StructuredLog.warn` rather than baked into the message string, and `redaction.batch` gains a `parallel` field — with the naming bug that was the second half of "no way to tell whether fan-out ran".

Not changed: the single-entry-huge-batch case still forks for zero parallelism. It costs one fork and the entry is genuinely large, so I'd rather leave it than add a second condition to a gate that has now been wrong twice.

Gate: **597 tests, 0 failures**, PMD clean.

Four rounds on this PR, four genuine defects — three of them mine in the fixes rather than the original change. Worth recording that the original serial loop would have been safer than a carelessly parallelised one.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,21 +126,131 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+    /**
+     * The content {@link #redactEntry} actually rewrites, which is what its cost scales with.
+     *
+     * <p>Counting only the {@code *MessagesJson} pair was wrong twice over. Those two are populated by
+     * the OTLP mapper alone — every pull adapter and the upload parser use the back-compat constructor
+     * and pass null — so the total was 0 for those sources and the parallel gate could never open there
+     * at any configured value. Even on OTLP it ignored {@code input}/{@code output}, which carry inline
+     * media. {@code InProcessSpool.measure} and {@code OtlpIngestService} already sum the same set.
+     */
+    private static long redactableBytes(RawEntry e) {
```
nidhi-nair, line 238: **Warning (Correctness/Observability)**: this counts 4 of the 5 things `redactEntry` rewrites — `metadata` is still missing, and `redactMetadata` (`:308-323`) regexes every non-usage string value in it.

`OtlpSpanMapper:89-90` builds `attrs` as the resource attributes plus every flattened span attribute and passes it in as `metadata` at `:159`; `gen_ai.input.messages`/`output.messages` are read out of it at `:135-136` and left in it, and the duplicate is only stripped later by `SpanBatchWriter` (this file's `redact` javadoc, `:287`). So on a normal LLM span the real work is ~2x what `bytes` reports, and a span whose payload sits in custom attributes (`input`/`output` null out of `firstNonBlank`, `OtlpSpanMapper:149-150`) counts **0 bytes** and can never fork at any `parallelism` — the same shape as the blocker closed last round, one field along.

Not a blocker (it fails safe to serial), but the javadoc at `:214` claims this is "the content `redactEntry` actually rewrites", which is not true as written.

**Fix:** add the metadata strings, reusing the predicate `redactMetadata` already applies:
```java
if (e.metadata() != null) {
    for (Map.Entry<String, Object> kv : e.metadata().entrySet()) {
        if (kv.getValue() instanceof String s && !GenAiAttributes.isUsageOrCostKey(kv.getKey())) n += s.length();
    }
}
```
or state the exclusion in the javadoc instead of claiming completeness.

nidhi-nair (commented):
## Code Review Summary

**Third review, of `5154970` on top of `0ca5b27`.**

**All prior blockers are closed. I verified each against the code and the JDK source rather than taking the response.**

- **Byte gate dead on non-OTLP sources — closed.** `redactableBytes` (`:222`) now sums `input + output + inputMessagesJson + outputMessagesJson`, the same set `OtlpIngestService:159` and `InProcessSpool.measure` already use. A pull/upload entry that counted 0 now counts its blob. (Worth noting for the record: `SubstrateWriter.enqueue` has exactly one production caller today, `OtlpIngestService:161`, so the non-OTLP half of that blocker was latent rather than live — but the OTLP half, inline media in `input`/`output` being uncounted, was live, and the fix is right either way.)
- **Thread naming — closed and correct.** `AtomicInteger seq` is a constructor local captured by the factory lambda; the pool holds the factory, so the counter outlives the constructor. Names are now genuinely distinct. See suggestion 2 for the one caveat.
- **`parallel` field on `redaction.batch` — added**, but the expression does not match the gate. See warning 2.
- **Interrupt no longer duplicates the work — closed, and the loop is correct.** I checked the spin question in the JDK 25 source: `ForkJoinTask.get` tests `Thread.interrupted()` at `:1017`, and `awaitDone` at `:425` does the same before setting `interrupts = -1`. Both **clear** the flag, so each `InterruptedException` leaves the loop's next `get()` free to park rather than re-throw immediately — no hot spin under `SubstrateWriter.shutdown`'s single interrupt, and a caller would have to re-interrupt in a tight loop to make it burn. The caller's `finally { Thread.currentThread().interrupt(); }` is right for exactly that reason: the flag was cleared by the throw, and it needs restoring on the normal return *and* on the `Error` propagation, which `finally` gives you. The `Error` rethrow inside the helper neither duplicates nor conflicts with the outer `catch (ExecutionException)` — once control is inside the sibling `catch (InterruptedException)` block, the outer catch is out of scope and cannot see anything thrown from it, so the helper has to handle it itself. Four duplicated lines, not duplicated behaviour.
- **`Error` rethrow, categorical WARN, `@PreDestroy`, javadoc — all still correct** after the reshuffle; `StructuredLog.warn(..., Markers.OPS, ...)` at `:195` now carries the fields as fields, and the DEBUG detail line is unchanged and still cannot egress.

What is left is three warnings — one of them the same bug class as the blocker just closed, one field further in — and no blockers.

Found 0 blockers and 3 warnings.

### Warnings

**1. `RedactionService.java:222` — (Correctness/Observability) `redactableBytes` counts 4 of the 5 things `redactEntry` rewrites; `metadata` is still missing, and on this path metadata is the whole attribute map.**

`redactEntry:258-274` rewrites `input`, `output`, **`metadata` via `redactMetadata`**, and both `*MessagesJson`. `redactMetadata:308-323` runs `RedactionEngine.applyToJson` over *every* non-usage string value in the map — real regex, per entry.

And `metadata` is not a few tags. `OtlpSpanMapper:89-90` builds `attrs` as the resource attributes plus every flattened span attribute, and passes it straight in as `metadata` at `:159`. `gen_ai.input.messages` / `gen_ai.output.messages` are *read out of* `attrs` at `:135-136` and **left in it**; this file's own `redact` javadoc at `:287` says `SpanBatchWriter` strips the attribute that duplicates a promoted column — downstream of redaction, so at `redactEntry` time the duplicate is present.

Two concrete consequences:

- **Common LLM span:** the real work is roughly double what `bytes` reports, so a 16 KiB gate effectively opens at ~32 KiB of scanned text, and the `bytes` field on `redaction.batch` — the number this PR argues is *the* cost model — under-reports by the same factor.
- **A batch that can never fork at any `parallelism`:** a span carrying its payload in custom attributes rather than `gen_ai.*` — a tool, retrieval or workflow span whose `input`/`output` come back null from `firstNonBlank` at `OtlpSpanMapper:149-150` — counts **0 bytes** while `redactMetadata` regexes the entire attribute map. That is structurally the blocker closed last round, moved one field along.

Not a blocker: the failure mode is *stays serial*, i.e. exactly the pre-PR behaviour, never wrong output. But the javadoc at `:214` states the sum is "the content `redactEntry` actually rewrites", and that sentence is not true as written.

**Fix:** add the metadata strings to the sum, reusing the predicate `redactMetadata` already applies —

```java
if (e.metadata() != null) {
    for (Map.Entry<String, Object> kv : e.metadata().entrySet()) {
        if (kv.getValue() instanceof String s && !GenAiAttributes.isUsageOrCostKey(kv.getKey())) n += s.length();
    }
}
```

— or, if you would rather keep the sum O(1) per entry, say in the javadoc that metadata is deliberately excluded and what that costs, instead of claiming completeness.

**2. `RedactionService.java:138` — (Observability) the `parallel` field does not report whether the fan-out ran.**

The gate is `pool == null || pool.isShutdown() || bytes < PARALLEL_THRESHOLD_BYTES` (`:162`); the field is `pool != null && bytes >= PARALLEL_THRESHOLD_BYTES`. It omits `isShutdown()`, so on the one path that guard exists for it logs `parallel=true` for a batch that ran serially. It also reports `true` for a batch that forked and then fell back through the `ExecutionException` catch at `:204` (that one at least has its own WARN). Since the whole point of adding the field was that there was no way to tell which path ran, a copy of the condition that is already a superset of it is the wrong shape.

**Fix:** report the path rather than re-derive it — have `redactAll` record it (return a small record, or take a `boolean[]`/`AtomicBoolean` out-param) and log that value.

**3. `RedactionService.java:162-166` — (Correctness) `isShutdown()` narrows the shutdown race, it does not close it.**

`shutdownPool()` can run between the check at `:162` and the submit at `:165`, and `ForkJoinPool` throws `RejectedExecutionException` on submit-after-shutdown (JDK 25 `ForkJoinPool.java:2611`). So the escape the comment at `:159-161` says it prevents is still reachable — out of a method whose javadoc at `:114-116` now enumerates `Error` as the one thing it propagates. The window is small and the consequence unchanged from before the guard (`drainLoop`'s `catch (RuntimeException)` drops the batch), so this is a warning, not a blocker — but the guard reads as a fix and is a narrowing.

**Fix:** keep the guard, and make it total with the catch it was standing in for:

```java
try {
    ForkJoinTask<List<RawEntry>> task = pool.submit(...);
    ...
} catch (RejectedExecutionException e) {
    return serially(entries, rules);
}
```

### Suggestions

- `RedactionService.java:238-243` — the helper's non-`Error` `ExecutionException` branch falls back to `serially` **silently**, while the identical case on the main path emits `redaction.parallel.failed` at `:195`. A fan-out failure that happens to coincide with a shutdown interrupt leaves no trace at all. Extract the shared handling so both paths log it.
- `RedactionService.java:99` — the counter is correct and the names are unique, but it counts worker *creations*, not slots: a non-common `ForkJoinPool`'s workers time out when idle and are recreated, so over a long-running process a 4-thread pool drifts to `redaction-137` and a thread dump no longer shows the bound. `"redaction-" + (seq.getAndIncrement() % parallelism)` keeps the names inside the configured width and repeats one only across a replacement — which is the thing you would want to see.
- `devdocs/reference/config-keys.md:42` — the missing period from the last round is still there: *"…reconciled onto every project by name  `parallelism` (default 4) is how many threads…"*.
- `RedactionProperties.java:47` — *"Above the core count it only adds contention"* still reads as contradicting the shipped default of 4 against the reference box, whose backend is pinned to 2 vCPU (`docker-compose.bench.yml:38`, with the box's budget spelled out at `:11-20` — note that file lives on the `docs/bench-harness` branch and is not reachable from this one, so the number is not re-derivable from this PR's tree). You offered to reword; one clause saying the 2:1 oversubscription is deliberate and is what the 2.14x measured would close it, and costs nothing.
- `RedactionService.java:234` — for completeness, not as a finding: if the pool is ever *stopping* while the waiter is interrupted, `awaitDone` cancels the task itself (JDK 25 `ForkJoinTask.java:426-433`) and `get()` then throws `CancellationException` **unwrapped**, past both catches in `awaitUninterruptibly`. Unreachable today — `shutdownPool` calls `shutdown()`, never `shutdownNow()`, and a running task keeps the pool non-quiescent — so no change is warranted; noting it only so it is not rediscovered as a mystery if that ever becomes `shutdownNow()`.

### Would I ship this?

Yes, with warning 1 taken first — it is a handful of lines and it is the one that keeps a real batch shape from ever forking. Warnings 2 and 3 are one-liners that belong in the same pass. **Every remaining defect fails safe to the pre-PR serial path**, so if this had to go today as it stands I would not block it; nothing here can produce wrong output, lose an entry, or leak content. The three fixes from last round all landed and all do what they claim this time, and the two I was most likely to find wrong — the uninterruptible loop and the `finally` that restores the flag — are correct for the right reasons.

**Verdict: changes-requested.**

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,21 +126,131 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
-            bytes += length(e.inputMessagesJson()) + length(e.outputMessagesJson());
-            out.add(redactEntry(e, rules));
+            bytes += redactableBytes(e);
         }
+        List<RawEntry> out = redactAll(entries, rules, bytes);
         StructuredLog.info(log, Markers.OPS, "redaction.batch")
                 .field("entries", entries.size())
                 .field("rules", rules.size())
                 .field("bytes", bytes)
+                .field("parallel", pool != null && bytes >= PARALLEL_THRESHOLD_BYTES)
```
nidhi-nair: **Warning (Observability)**: this does not match the gate. `redactAll:162` is `pool == null || pool.isShutdown() || bytes < T`; this is `pool != null && bytes >= T`. It omits `isShutdown()`, so on the one path that guard exists for it logs `parallel=true` for a batch that ran serially — and it also reports `true` for a batch that forked and fell back through the `ExecutionException` catch at `:204`.

The field was added because there was no way to tell which path ran; a re-derived superset of the condition doesn't give you that. **Fix:** have `redactAll` report the path it took (return it, or set an `AtomicBoolean`/`boolean[]` out-param) and log that value.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,21 +126,131 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+     * and the parallelism stays the stated bound.
+     *
+     * <p>Small batches stay on the calling thread: below {@link #PARALLEL_THRESHOLD_BYTES} the fork and
+     * join cost more than the regex saves.
+     */
+    private List<RawEntry> redactAll(List<RawEntry> entries, List<CompiledRule> rules, long bytes) {
+        // isShutdown, because @PreDestroy can close the pool while a drainer still holds a claimed batch
+        // (SubstrateWriter.shutdown does not join them) and submit would then throw
+        // RejectedExecutionException out of a method documented as never throwing.
+        if (pool == null || pool.isShutdown() || bytes < PARALLEL_THRESHOLD_BYTES) {
```
nidhi-nair, line 164: **Warning (Correctness)**: `isShutdown()` narrows the race, it does not close it. `shutdownPool()` can run between this check and the `pool.submit(...)` three lines down, and `ForkJoinPool` throws `RejectedExecutionException` on submit-after-shutdown (JDK 25 `ForkJoinPool.java:2611`). So the escape the comment above says it prevents is still reachable, out of a method whose javadoc at `:114-116` now enumerates `Error` as the one thing it propagates.

Small window, and the consequence is unchanged from before the guard (`drainLoop`'s `catch (RuntimeException)` drops the batch) — hence warning, not blocker. **Fix:** keep the guard and make it total:
```java
try {
    ForkJoinTask<List<RawEntry>> task = pool.submit(...);
    ...
} catch (RejectedExecutionException e) {
    return serially(entries, rules);
}
```

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -79,21 +126,131 @@ public List<RawEntry> redactBatch(String projectId, List<RawEntry> entries) {
...
+                return task.get();
+            } catch (InterruptedException ignored) {
+                // Deliberately swallowed here and re-raised by the caller's finally: the point of this
+                // loop is to finish work the pool is already doing rather than abandon and repeat it.
+            } catch (ExecutionException ee) {
+                if (ee.getCause() instanceof Error err) {
+                    err.addSuppressed(ee);
+                    throw err;
+                }
+                return serially(entries, rules);
```
nidhi-nair, line 276: **Suggestion (Consistency)**: this branch falls back to `serially` **silently**, while the identical non-`Error` case on the main path emits `redaction.parallel.failed` at `:195`. A fan-out failure that coincides with a shutdown interrupt therefore leaves no trace at all. Extract the shared handling (log + `serially`) so both paths report it.

backend/substrate/src/main/java/ai/tessary/redaction/RedactionService.java
```diff
@@ -56,7 +62,46 @@ public class RedactionService {
...
+        // says nothing about which subsystem is burning the core. A counter rather than getPoolIndex():
+        // the pool assigns that index in registerWorker, AFTER this factory returns, so naming from it
+        // labels every worker "redaction-0" and buys nothing.
+        AtomicInteger seq = new AtomicInteger();
+        this.pool = parallelism > 1
+                ? new ForkJoinPool(
+                        parallelism,
+                        p -> {
+                            ForkJoinWorkerThread t = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(p);
+                            t.setName("redaction-" + seq.getAndIncrement());
```
nidhi-nair, line 100: **Suggestion (Observability)**: the capture is correct — `seq` is a constructor local, effectively final, kept alive by the factory the pool holds — and the names are genuinely unique now. One caveat: it counts worker *creations*, not slots. A non-common `ForkJoinPool`'s workers time out when idle and are recreated, so over a long-running process a 4-thread pool drifts to `redaction-137` and a thread dump no longer shows the configured bound. `"redaction-" + (seq.getAndIncrement() % parallelism)` keeps names inside the width and repeats one only across a replacement, which is the thing worth seeing.

PR #7, opened 2026-09-09
fix(db): the Hikari pool binds spring.datasource.hikari.*, which it had been ignoring

nidhi-nair:
## The bug

`DataSourceConfig` builds the pool programmatically, which replaces Boot's auto-configured `DataSource` — so the entire `spring.datasource.hikari` block in `application-production.yaml` was **inert**. Every value in it was decoration:

```yaml
spring:
  datasource:
    hikari:
      maximum-pool-size: 10      # matched the hardcoded value by coincidence
      minimum-idle: 2            # code used 1
      connection-timeout: 10000  # code used Hikari's 30s default
      pool-name: evals-hikari    # code used the same literal
```

An operator raising `maximum-pool-size` to relieve pool pressure — the exact lever you reach for when the pool is the ceiling — would have seen no effect at all.

## Why it mattered in practice

`connection-timeout` is the one that had drifted. The yaml asks for 10 s; the code was on Hikari's 30 s default. An overload measurement showed precisely that: **p99 of 34 s and a worst case of 55 s**, where the documented behaviour was a 10 s failure. Under Loom the pool, not the thread count, is the write path's ceiling — the file's own comment says so — so how long a caller waits for a connection is the difference between backpressure and a stall.

This surfaced while measuring ingest capacity: pool exhaustion under overload produced `CannotGetJdbcConnectionException` → **HTTP 500**, and OTLP clients retry 503 but not 500, so those spans were lost rather than resent.

## The fix

`@ConfigurationProperties("spring.datasource.hikari")` on the bean method, so the yaml binds onto the pool the code builds. The values in code become defaults that production overrides.

Built with the **no-argument** `HikariDataSource` constructor on purpose: `HikariDataSource(HikariConfig)` seals its configuration immediately, and `@ConfigurationProperties` binds *after* the factory method returns — binding into a sealed pool throws. That is the one non-obvious part of the change.

Also: the pool name stops saying `evals`.

## Scope and risk

- **This changes production behaviour**, deliberately: `minimum-idle` becomes 2 (was 1) and `connection-timeout` becomes 10 s (was 30 s). Both are what the yaml has been asking for. The shorter timeout means a saturated pool fails faster, which is the intent — but it is a real behaviour change on the write path and worth a reviewer's eye rather than being waved through as a config tidy-up.
- No test-side `spring.datasource.*` config exists to collide with (checked).
- `scripts/check-backend.sh`: **597 tests, 0 failures**.

Found while measuring ingest capacity; the 34 s p99 above is from that run.


backend/app/src/main/resources/application-production.yaml
```diff
@@ -27,7 +30,7 @@ spring:
       minimum-idle: 2
       connection-timeout: 10000
       leak-detection-threshold: 30000
-      pool-name: evals-hikari
+      pool-name: tessary-hikari
```
nidhi-nair: **Blocker (Consistency)**: this rename reverses a decision recorded one commit ago. `devdocs/reference/telemetry-naming.md:95-98` states the Hikari pool name was deliberately NOT renamed because it is a Micrometer `pool=` tag and renaming splits the series, and commit `f130c79` reverted exactly this line back to `evals-hikari` to make that true. This PR re-applies it as a drive-by with no mention of that decision and no doc update. Drop the rename from this PR - it is unrelated to the binding fix. If you want it, give it its own change that rewrites that telemetry-naming paragraph. The same applies to `DataSourceConfig.java:65`.

backend/core/src/main/java/ai/tessary/db/DataSourceConfig.java
```diff
@@ -27,30 +27,46 @@
...
+        HikariDataSource ds = new HikariDataSource();
+        ds.setDriverClassName("org.postgresql.Driver");
+        ds.setMaximumPoolSize(10);
+        ds.setMinimumIdle(1);
+        ds.setLeakDetectionThreshold(30_000);
+        ds.setPoolName("tessary-hikari");
+        ds.setJdbcUrl(jdbcUrl);
+        ds.setUsername(props.getDbUsername());
+        ds.setPassword(props.getDbPassword());
         log.info("postgres datasource at {}", jdbcUrl);
```
nidhi-nair, line 72: **Warning (Correctness)**: this now logs a value that may not be what the pool connects with. `@ConfigurationProperties` binds after this method returns, so `spring.datasource.hikari.jdbc-url` (relaxed-bound from `SPRING_DATASOURCE_HIKARI_JDBC_URL`) overrides `TESSARY_JDBC_URL` - while this line and the fail-fast guard at line 52 both still report the `TESSARY_*` value. The one line telling an operator which database the process is on can be silently wrong. Fix: log after binding (an `InitializingBean`/`@PostConstruct` reading `ds.getJdbcUrl()`), or bind a narrower pool-tuning object so the URL and credentials keep a single source.

nidhi-nair (commented):
## Code Review Summary

Found 1 blocker and 2 warnings. The core fix is correct and I verified every claim it makes; the problem is a drive-by rename riding along with it.

**What I verified (all three claims hold):**

1. **The yaml really was inert.** `DataSourceConfig` declares a `DataSource` bean, so Boot's `DataSourceAutoConfiguration` (`@ConditionalOnMissingBean(DataSource.class)`) backs off entirely — and with it `DataSourceConfiguration.Hikari`, which is the only thing that carries `@ConfigurationProperties("spring.datasource.hikari")`. `DataSourceProperties` never gets created either. Nothing else in the tree reads those keys: the only consumers of the bean are `LiquibaseConfig:30` and `DetectionTableSchemaCheck:40`, both by `DataSource`. `spring.datasource` appears in no other yaml, no test resource, and no compose/env file (`docker-compose.yml` passes only `TESSARY_*` and `SPRING_PROFILES_ACTIVE`). Claim confirmed.

2. **The no-arg-constructor reasoning is exactly right.** Checked against HikariCP 7.0.2 sources: `HikariDataSource(HikariConfig)` runs `copyStateTo(this)`, starts the pool, then `this.seal()` — and every `HikariConfig` setter calls `checkIfSealed()`. The no-arg form leaves `fastPathPool == null` and seals only inside the first `getConnection()`. `ConfigurationPropertiesBindingPostProcessor` binds in `postProcessBeforeInitialization`, i.e. after the factory method returns and long before any connection is taken. So binding into the old shape would indeed have thrown, and the new shape binds cleanly. Claim confirmed.

3. **10s connection-timeout is safe here.** `connectionTimeout` bounds *acquisition from the pool*, not query or transaction duration — a long-running batch, retention sweep or rollup is unaffected once it holds a connection. Startup/Liquibase is fine: `docker-compose.yml:266-268` gates `backend` on `postgres: service_healthy`, so the DB is accepting before the pool ever starts, and 10s is generous for a docker-network connect. Background demand on the pool is small by construction — every `AsyncConfig` executor is concurrency-limited to 2 — and failures are survivable: `TraceRollupWorker:98,143` catches per-unit `RuntimeException` and the job queue is `FOR UPDATE SKIP LOCKED` re-claimable; `RetentionSweeper` is `@Scheduled(fixedDelay 1h)` and simply re-runs. The blast radius of the shorter timeout is confined to the saturation case on the ingest write path, which is precisely where failing fast is the intent. No objection to this part.

**On exposure (question 4):** binding the whole prefix does mean `spring.datasource.hikari.jdbc-url` / `username` / `password` / `read-only` now take effect. That is not a privilege escalation — anyone who can set `SPRING_DATASOURCE_HIKARI_*` on the container already sets `TESSARY_JDBC_URL` — so it is not a security finding. It does introduce one concrete correctness defect, below.

**On the return type (question 5):** no consequence. Nothing in the tree injects `HikariDataSource`; both consumers take `DataSource`, and the bean stays assignable. One latent trap worth knowing: `SubstrateWriteIntegrationTest:89` replaces the bean with a JDK proxy that is *not* a `HikariDataSource` while the declared bean type now says it is — harmless today because nothing resolves by the concrete type, but it will bite whoever first writes `getBean(HikariDataSource.class)`.

### Blockers

- `backend/app/src/main/resources/application-production.yaml:33` and `backend/core/src/main/java/ai/tessary/db/DataSourceConfig.java:65` — **(Consistency)** The `evals-hikari` → `tessary-hikari` rename reverses a decision taken one commit ago and recorded in the repo. `devdocs/reference/telemetry-naming.md:95-98` says, in as many words, that the Hikari pool name was deliberately *not* renamed because it is a Micrometer `pool=` tag and renaming it splits the series; commit `f130c79` ("address crew review on the tessary namespace rename") reverted exactly this pair of lines back to `evals-hikari` to make that true. This PR re-applies the rename as an aside ("Also: the pool name stops saying `evals`"), does not mention the recorded decision, and does not update the doc — which the same doc's own rule and AGENTS.md's same-PR co-update rule both require. Fix: drop the rename from this PR. It is unrelated to the binding fix, which stands on its own. If the rename is genuinely wanted, it belongs in its own change that rewrites that telemetry-naming paragraph and states what happens to the existing series.

### Warnings

- `backend/core/src/main/java/ai/tessary/db/DataSourceConfig.java:69` — **(Correctness)** `log.info("postgres datasource at {}", jdbcUrl)` now reports a value that is not necessarily what the pool connects with. Binding happens after this method returns, so `spring.datasource.hikari.jdbc-url` (relaxed-bound from `SPRING_DATASOURCE_HIKARI_JDBC_URL`) wins over `TESSARY_JDBC_URL` — while both this log line and the fail-fast guard at `:52` still describe the `TESSARY_*` value. The one line that tells an operator which database the process is on can now be wrong, silently. Fix: log after binding — move the line into a small `InitializingBean`/`@PostConstruct` that reads `ds.getJdbcUrl()` — or, if the intent is that `TESSARY_JDBC_URL` remains the only source of the URL, bind a narrower pool-tuning properties object instead of the whole prefix.

- `devdocs/reference/config-keys.md` — **(Docs)** `spring.datasource.hikari.*` goes from decoration to live, operator-facing configuration in this PR, and AGENTS.md names this file the config-key map. It has no entry for the prefix. Add one row saying the prefix binds onto `DataSourceConfig`'s pool and that its `jdbc-url`/`username`/`password` members override `TESSARY_*` (or that they must not be set, if you take the narrowing fix above).

### Suggestions

- `backend/core/src/main/java/ai/tessary/db/DataSourceConfig.java:31-41` — the two new javadoc paragraphs narrate the incident (34s p99, 55s worst case, "which until now it silently ignored"). The repo's docs are current-state, and AGENTS.md limits comments to non-obvious invariants. The constructor comment at `:57-59` clearly earns its place; these two do not and will read as stale archaeology in a month. One sentence — "the yaml binds onto this pool and overrides these defaults" — carries everything a future reader needs.
- `backend/app/src/main/resources/application-production.yaml:18-20` — restates the javadoc rather than linking to it ("one home per fact"). And directly beneath it, lines 26-28 still explain the pool size in terms of "the grading workers" and "each unit's graders", which Track A deleted; the PR rewrites this block's header without fixing the stale sentence under it.

**Note for the human merger:** this is a deliberate production-behaviour change in `application-production.yaml` (`minimum-idle` 1→2, `connection-timeout` 30s→10s). I judge the 10s timeout safe for every caller for the reasons above, but the merge call on a prod-config change is a human's.

Verdict: changes-requested.

PR #8, opened 2026-09-09
docs(bench): check in the ingest load harness the capacity numbers came from

nidhi-nair:
## Why

`devdocs` quotes ingest capacity figures — the drain ceiling in the runbook's objectives, the 2.14× on `tessary.redaction.parallelism`, the token cache's latency numbers — and **nothing in the tree could re-derive any of them**. A capacity figure with no way to re-run it is a claim, not a measurement, which is the standard the classifier work is already held to.

## What lands

`scripts/bench/` — the generator and three drivers:

| | |
|---|---|
| `otlp-load.js` | **Open-loop** at a fixed arrival rate. That is the load-bearing choice: a closed-loop generator slows down exactly when the server does, so it can never observe a rejection rate at a stated offered load. Hand-encodes protobuf the way `scripts/emit-span.js` does, so it needs nothing installed. |
| `projects.sh` | Mints N projects and tokens. The spool partitions by project, so a single-project run measures a single drainer whatever else is configured — an easy way to measure nothing by accident. |
| `ramp.sh` | Steps offered load; records accepted, refused, **rows actually counted out of Postgres**, and container CPU per step. |
| `spike.sh` | Prices burst headroom by the second the first `503` lands. |
| `mem-probe.sh` | cgroup, heap and NMT at a stated load. |

`docker-compose.bench.yml` pins the reference box (4 vCPU / 8 GiB; backend 2 vCPU / 2 GiB — which is what makes `-XX:MaxRAMPercentage=75` mean a 1.5 GiB heap rather than a share of whatever the host happens to have). Opt-in: nothing loads it unless named with `-f`.

## Two things the README records because nobody would guess them

- **Reproducing an A/B means toggling a property, not swapping an older image in as the control.** An older build cannot boot against a forward-migrated schema — its Liquibase changelog no longer matches. I lost a run to that.
- **`mem-probe.sh` needs a sidecar JDK sharing the backend's PID namespace *and* its `/tmp`.** The shipped image is a JRE with no `jcmd`, and HotSpot's attach handshake goes through a socket in the target's `/tmp` that a sidecar in its own mount namespace cannot see. That is the only reason the overlay mounts `/tmp` as a shared volume.

## Notes for review

- The README is explicit that these numbers **do not transfer to the hosted deployment** — that is a burstable `t3.small`, and the CPU-credit behaviour alone breaks the comparison.
- Container names (`tessary-postgres-1`) are hardcoded to the default compose project name; DB user/name come from `POSTGRES_USER`/`POSTGRES_DB` with the compose defaults.
- Nothing here runs in `task check` — it needs a live stack and takes minutes per ladder. It is a tool, not a gate.
- **Judgement call worth flagging:** I landed the generator and the general drivers but not the bespoke A/B wrappers from the measuring session, on the grounds that the README's documented commands cover them and `AGENTS.md` discourages demo code. If you would rather have the exact scripts that produced the published tables, say so and I will add them.

### DIFF
--- a/docker-compose.bench.yml
+++ b/docker-compose.bench.yml
@@ -1,0 +1,73 @@
+# Ingest throughput measurement overlay — layered on docker-compose.yml (the SELF-HOST reference
+# stack, not the dev stack). Opt-in: nothing loads it unless you name it with -f.
+#
+# It is checked in so the capacity figures quoted in devdocs can be re-derived rather than taken on
+# trust — see scripts/bench/README.md for what to run and how to read the output.
+#
+#   docker compose -f docker-compose.yml -f docker-compose.bench.yml up -d
+#
+# It does two things and nothing else.
+#
+# 1. PINS THE REFERENCE BOX. Docker Desktop hands this stack the whole VM (12 CPUs / 19.5 GiB here),
+#    which is not a machine anyone deploys on, and a throughput number measured against it is a
+#    number about my laptop. The limits below carve out a stated 4 vCPU / 8 GiB box:
+#
+#      backend         2.00 vCPU   2048 MiB   (mem_limit is docker-compose.yml's own declared 2g)
+#      postgres        1.25 vCPU   3072 MiB
+#      frontend/caddy  0.50 vCPU    512 MiB
+#      sandbox-runner  0.25 vCPU   1536 MiB   (declared; idle throughout, it runs no ingest work)
+#      ------------------------------------
+#      total           4.00 vCPU   7168 MiB   leaving ~1 GiB for the host and its page cache
+#
+#    Postgres gets its own limit because without one the shared page cache grows into whatever the
+#    VM has, and a drain rate measured against a fully-cached working set is not the rate a real box
+#    sustains once the tables outgrow RAM.
+#
+# 2. TURNS ON JFR AND NATIVE MEMORY TRACKING, because the heap is the thing most likely to be
+#    misread here. -XX:MaxRAMPercentage=75 in backend/Dockerfile sizes the heap off the container
+#    limit, so the 2g above is what makes the heap 1.5 GiB rather than whatever share of 19.5 GiB
+#    the ergonomics would otherwise pick. NMT is what separates that heap from the metaspace, thread
+#    stacks, code cache and direct buffers the cgroup limit also has to cover; it costs a few percent
+#    of footprint itself, which is why it is here and not in the shipped file.
+#
+# JAVA_TOOL_OPTIONS is additive to the image ENTRYPOINT's flags, so the production heap, OOM-dump
+# and OOM-exit settings all still apply exactly as shipped.
+
+services:
+  backend:
+    cpus: 2.0
+    environment:
+      JAVA_TOOL_OPTIONS: >-
+        -XX:NativeMemoryTracking=summary
+        -XX:StartFlightRecording=name=ingest,settings=${BENCH_JFR_SETTINGS:-default},disk=true,maxsize=1g,maxage=2h,dumponexit=true,filename=/jfr/ingest.jfr
+        -XX:FlightRecorderOptions=repository=/jfr/repo
+      PYROSCOPE_AGENT_ENABLED: "false"
+      # The heartbeat is the queue-depth instrument, and once a minute is too coarse to see a step
+      # of a ramp. Cadence only — it changes what is logged, never what is admitted or drained.
+      TESSARY_INGEST_SUBSTRATE_REPORT_MS: "10000"
+      # Passthrough so an A/B can flip one property between arms (redaction parallelism, the token
+      # cache) without editing this file. Empty in a normal run, which leaves every default in place.
+      SPRING_APPLICATION_JSON: ${SPRING_APPLICATION_JSON:-}
+    volumes:
+      - ./.local/bench/jfr:/jfr
+      # The shipped image is a JRE: it has java and jfr but no jcmd, so there is no way to ask the
+      # live JVM for an NMT summary or a heap info. A sidecar JDK sharing this container's PID
+      # namespace can, but HotSpot's attach handshake goes through a socket in the target's /tmp,
+      # which a sidecar in its own mount namespace cannot see. Making /tmp a named volume both
+      # containers mount is what lets the handshake complete. Storage only; nothing about the
+      # application changes.
+      - bench-tmp:/tmp
+
+  postgres:
+    cpus: 1.25
+    mem_limit: 3g
+
+  frontend:
+    cpus: 0.5
+    mem_limit: 512m
+
+  sandbox-runner:
+    cpus: 0.25
+
+volumes:
+  bench-tmp:
--- a/scripts/bench/README.md
+++ b/scripts/bench/README.md
@@ -1,0 +1,93 @@
+# Ingest load harness
+
+Reproduces the ingest capacity numbers quoted in
+[`devdocs/guides/ingest-runbook.md`](../../devdocs/guides/ingest-runbook.md) § *The objectives* and in
+the `tessary.redaction.parallelism` / `tessary.auth.token-cache.*` entries of
+[`devdocs/reference/config-keys.md`](../../devdocs/reference/config-keys.md).
+
+It exists because those numbers were quoted before anything here could re-derive them. A capacity
+figure with no way to re-run it is a claim, not a measurement.
+
+## What it is
+
+| File | What it does |
+|---|---|
+| `otlp-load.js` | Open-loop OTLP/HTTP generator. Hand-encodes protobuf (same approach as `scripts/emit-span.js`) so it depends on nothing but Node. |
+| `projects.sh` | Mints N projects in one org and writes one write-scoped token per line to `tokens.txt`. |
+| `ramp.sh` | Steps offered load up a ladder, recording accepted/refused, rows actually written, and container CPU per step. |
+| `spike.sh` | Holds a baseline rate, applies a burst, and records the second the first `503` appeared. |
+| `mem-probe.sh` | cgroup, heap and NMT figures at a stated load. Needs a sidecar JDK — see below. |
+| `../../docker-compose.bench.yml` | Pins the reference box and turns on JFR + native memory tracking. |
+
+**Open-loop is the point.** A closed-loop generator (N workers each waiting for its response) slows
+down exactly when the server does, so it can never observe a rejection rate at a stated offered load.
+
+## Running it
+
+```bash
+# 1. Bring the stack up on the pinned reference box (4 vCPU / 8 GiB; backend 2 vCPU / 2 GiB)
+HTTP_PORT=8000 docker compose -f docker-compose.yml -f docker-compose.bench.yml up -d
+
+# 2. Mint projects + tokens (1 for single-project runs, 8 for anything about drainers)
+cd scripts/bench && bash projects.sh 1
+
+# 3. A rate ladder
+bash ramp.sh myrun 180 400 460 500
+
+# 4. Burst headroom from a 420 spans/s baseline
+bash spike.sh myburst 420 30 2 3 4 6
+```
+
+`otlp-load.js --tokens tokens.txt` round-robins across projects; `--token <one>` pins a single one.
+That distinction matters: the spool partitions by project, so a single-project run measures a single
+drainer no matter what else is configured.
+
+## Reproducing the published comparisons
+
+Both were A/B'd by toggling one property and changing nothing else, with the substrate truncated
+between arms so index size does not confound the result.
+
+```bash
+# Redaction parallelism (config-keys.md: 476 -> 1021 spans/s, 2.14x)
+SPRING_APPLICATION_JSON='{"tessary":{"redaction":{"parallelism":1}}}' \
+  docker compose -f docker-compose.yml -f docker-compose.bench.yml up -d --force-recreate backend
+node otlp-load.js --token "$(head -1 tokens.txt)" --rate 1200 --spans-per-request 40 --duration 90 --warmup 15
+# then the same with parallelism 4
+
+# Token cache (the p50 61 ms -> 2.6 ms control)
+SPRING_APPLICATION_JSON='{"tessary":{"auth":{"token-cache":{"enabled":false}}}}' ...
+```
+
+Swapping an older *image* in as the control does not work: an older build cannot boot against a
+forward-migrated schema, because its Liquibase changelog no longer matches. Toggle the property.
+
+## Reading the results
+
+The generator prints a JSON summary; `--out` also writes a per-second JSONL. What each number means:
+
+- **`accepted_span_rate`** — spans the front door took. This is the capacity figure.
+- **`drain`** (rows counted straight out of Postgres) — what actually landed. It should track accepted;
+  a gap means the queue was still draining when the run ended.
+- **`refused`** — `503`s. Every one is a retryable answer the exporter will resend.
+- **`other`** — anything else, and **any non-zero value here is a finding**: `500` is not in OTLP's
+  retryable set, so those spans are lost rather than resent.
+
+Server-side, `event="ingest.throughput"` carries `queue_bytes`, `shed_batches` and
+`spool_oldest_age_ms`; the bench overlay drops its cadence to 10 s so a ladder step is visible.
+
+## `mem-probe.sh` needs a sidecar JDK
+
+The shipped backend image is a **JRE** — it has `java` and `jfr` but no `jcmd`, so the live JVM cannot
+be asked for a native-memory summary. A sidecar JDK sharing the backend's PID namespace can, but
+HotSpot's attach handshake goes through a socket in the target's `/tmp`, which a sidecar in its own
+mount namespace cannot see. The bench overlay mounts `/tmp` as a named volume both containers share,
+which is what completes the handshake. That mount exists for this and nothing else.
+
+JDK Mission Control is not required to read the results — the `jfr` CLI does it — but the dumped
+`.jfr` files open in JMC if you want its automated-analysis rules.
+
+## What the numbers are not
+
+Measured on a developer machine (arm64) against the self-host reference shape. Tessary's own hosted
+deployment is a burstable `t3.small`, so a sustained figure from here does not transfer to it — the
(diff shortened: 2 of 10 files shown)


scripts/bench/ramp.sh
```diff
@@ -0,0 +1,68 @@
...
+# heartbeat's queue depth, and container CPU/memory.
+#
+#   bash ramp.sh <label> <duration-per-step-seconds> <rate> [rate...]
+set -euo pipefail
+cd "$(dirname "${BASH_SOURCE[0]}")"
+ROOT="$(cd ../.. && pwd)"
+
+LABEL="$1"; shift
+STEP="$1"; shift
+TOKEN="$(cat token.txt)"
```
nidhi-nair: **Blocker (Correctness)**: reads `token.txt`, but `projects.sh` writes `tokens.txt`. The README quickstart (`projects.sh 1` then `ramp.sh`) dies here with `cat: token.txt: No such file or directory`. Same at `spike.sh:15`. Fix: `TOKEN="$(head -1 tokens.txt)"`, which is what README:54 already does.

nidhi-nair (commented):
> Verdict is **changes-requested**. Posted as a COMMENT review because GitHub does not allow requesting changes on your own PR.

## Code Review Summary

Found 3 blocker(s) and 11 warning(s), plus 6 suggestions.

The harness is the right idea and the reasoning in the comments is the most valuable thing in the diff — the open-loop rationale, the `/tmp` attach-socket explanation and the reference-box arithmetic are all things a future reader would otherwise have to rediscover. But **as checked in, two of the three drivers cannot run**: `ramp.sh` and `spike.sh` read `token.txt` while `projects.sh` writes `tokens.txt`, so the README's own four-step quickstart fails at step 3. And `projects.sh` itself aborts on any Linux box, which is the audience the README addresses.

**What I verified and confirms out** (so nobody re-checks it): the merged overlay renders — `docker compose -f docker-compose.yml -f docker-compose.bench.yml config` exits 0, `bench-tmp` resolves to `tessary_bench-tmp` matching `mem-probe.sh`, backend keeps its base `mem_limit: 2g`, `.local/` is already gitignored so the `/jfr` bind mount is safe, and `redpanda` is `profiles: ["kafka"]` so the box arithmetic does not have to account for it. No script calls a script that was not included (no `analyse.py`, `batch-sweep.sh` or `ab.sh` references) and there are no absolute authoring-machine paths anywhere. `bash -n` and `shellcheck -S warning` are clean on all four scripts; `node --check` is clean. The JRE/no-`jcmd` claim checks out (`backend/Dockerfile:75` is `eclipse-temurin:25-jre`, `MaxRAMPercentage=75` at `:135`). The Liquibase claim checks out and is stronger than the README states — `backend/core/src/main/resources/db/changelog/db.changelog-master.yaml` says outright that "a database that predates the baseline fails its first boot loudly and is recreated". The redaction numbers check out at `devdocs/reference/config-keys.md:42` (476 -> 1,021, 2.14x). `tessary.ingest.substrate.report-ms` and the `span` table both exist.

Answering the framing questions directly: **yes, it belongs in the repo** — an unmeasured capacity figure is exactly the liability AGENTS.md's closing section names, and `scripts/bench/` is the right home. It is adequately clear it is not part of `task check`. The problems are that it does not run, that the docs it exists to serve never point at it, and that one of the two published comparisons it claims to reproduce cites a number that is not published anywhere in the tree.

### Blockers

- **`scripts/bench/ramp.sh:14`** and **`scripts/bench/spike.sh:15`** — **(Correctness)** `TOKEN="$(cat token.txt)"`, but `projects.sh:14,28,33` writes **`tokens.txt`**. `README.md:32-35` runs `projects.sh 1` and then `ramp.sh`, which dies on `cat: token.txt: No such file or directory`. Neither driver works as checked in. Fix: read `tokens.txt` (`TOKEN="$(head -1 tokens.txt)"`, matching what `README.md:54` already does).
- **`scripts/bench/projects.sh:13`** — **(Correctness)** `mktemp -t tessary-bench-cookies` fails on GNU coreutils: `mktemp: too few X's in template 'tessary-bench-cookies'`, exit 1 (verified in `debian:12`). It only works because it was authored on macOS/BSD. Under `set -euo pipefail` the script aborts before it does anything, on every Linux self-host box — the exact shape `README.md:91` says the numbers were measured against. Fix: `mktemp -t tessary-bench-cookies.XXXXXX`, or just `mktemp`.
- **`scripts/bench/projects.sh:14,28,33`** — **(Security)** plaintext write-scoped bearer tokens are written to `scripts/bench/tokens.txt`, inside the working tree, and nothing gitignores it — `.gitignore` covers `.local/` and nothing under `scripts/bench/`. Same for the `results/` trees (`ramp.sh:15`, `spike.sh:16`). `.gitleaks.toml:19` says `tsy_` is a recognised secret format, so the likely outcome is a future contributor tripping the scanner, but the failure mode if they do not is a live token on `main`. Add `scripts/bench/tokens.txt` and `scripts/bench/results/` to `.gitignore` in this PR.

### Warnings

- **`docker-compose.bench.yml:59`, `scripts/bench/mem-probe.sh:19-20`** — **(Security)** `bench-tmp:/tmp` is a shared named volume, and `README.md:78-84` explains the mechanism but not the two consequences for a self-hoster who leaves the overlay on. (a) The HotSpot attach socket now lives somewhere any container mounting `tessary_bench-tmp` can reach, and the attach API includes `JVMTI.agent_load` — that is arbitrary code inside the backend JVM, and `mem-probe.sh` already demonstrates the attach as `--user root`. (b) A named volume is not wiped on container restart, so whatever the app spools to `/tmp` (multipart temp files, JFR repo spillover) accumulates on the host indefinitely. Add an explicit "this overlay is for a measuring session; do not leave it enabled" line to the README and the compose header. The same applies to leaving JFR + NMT on: the header at `:26-31` explains why NMT is not shipped, but never tells the reader to take the overlay back off.
- **`scripts/bench/README.md:57-58`** — **(Correctness)** "the p50 61 ms -> 2.6 ms control" is attributed to the `tessary.auth.token-cache.*` entry of config-keys. That entry (`devdocs/reference/config-keys.md:47`) carries "bcrypt was 62% of backend CPU on the ingest path" and no latency pair, and the strings `61 ms` / `2.6 ms` appear nowhere in the tree. The README is pointing at a published number that is not published. Either land the number in config-keys.md in this PR or cite the one that is actually there.
- **Docs policy** — **(Consistency)** nothing in `devdocs/` links to `scripts/bench/`. `devdocs/reference/config-keys.md:42,47` and `devdocs/guides/ingest-runbook.md:10-25` still carry the figures with no pointer to the harness, so a reader of the *owning* doc still cannot re-derive them, which is the PR's stated purpose. AGENTS.md's "same-PR co-update" wants that one-line link added here — one line in each owning doc, pointing at `scripts/bench/README.md`, and the link direction stays one-home-per-fact.
- **`scripts/bench/README.md:3-4`** — **(Consistency)** claims to reproduce "the drain ceiling in the runbook's objectives", but `devdocs/guides/ingest-runbook.md:17` states O2 as >= 200 spans/s and names `SubstrateWriteIntegrationTest` as where it comes from. Two docs now give two provenances for one number. Pick one and make the other link to it.
- **`scripts/bench/otlp-load.js:46,152`; `scripts/bench/ramp.sh:34,56`** — **(Correctness)** `--max-inflight` (default 512) silently turns the generator closed-loop under load: `send()` returns early and increments `skipped` without sending, so the offered rate stops being the stated rate exactly when the server is struggling. `skipped` is in the JSON summary but `steps.csv` never records it, so a step that was inflight-capped is indistinguishable from one that was not. Given `README.md:22` makes open-loop the load-bearing claim, `skipped` has to appear in the ladder table (and ideally be non-zero-checked the way `other` is at `README.md:71-73`).
- **`scripts/bench/README.md:68` vs `scripts/bench/ramp.sh:56`** — **(Consistency)** the README teaches `accepted_span_rate` as "the capacity figure", but `ramp.sh` records `measured_span_rate` and never emits `accepted_span_rate` at all. The headline number is missing from the tool that produces the published table.
- **`scripts/bench/mem-probe.sh:33-34`** — **(Correctness)** the heading says "heap, after a full GC (live set, not floating garbage)" but the command is `jcmd GC.heap_info`, which reports current occupancy and triggers nothing. As written the recorded "live set" includes floating garbage. Either run `GC.run` first or drop the parenthetical — the whole point of the file is that the memory numbers are trustworthy.
- **`scripts/bench/projects.sh:12`** — **(Correctness)** `B=http://localhost:8000` is not overridable, while `README.md:29` instructs the user to set `HTTP_PORT`. Any other port and every curl fails. Make it `B="${BENCH_BASE_URL:-http://localhost:8000}"`.
- **`scripts/bench/ramp.sh:18`, `scripts/bench/mem-probe.sh:19,27`** — **(Consistency)** `tessary-backend-1` / `tessary-postgres-1` / `tessary_bench-tmp` are only correct for the default compose project name, and `POSTGRES_USER`/`POSTGRES_DB` are read from the *host* shell while compose reads them from `.env` — a self-hoster who set them in `.env` without exporting gets the `tessary` fallback and a psql failure. The PR body says all this; the README does not, and the PR body does not ship. One paragraph in the README, please.
- **`scripts/bench/projects.sh:20,25,32`** — **(Correctness)** no error handling on the curl/jq chain. A failed signup makes `ORG` the literal string `null`, `mint` then appends `null` to `tokens.txt`, and the failure only surfaces minutes later as 401s inside a load run. A `[ "$ORG" != null ]` guard after line 20 and the same on `plaintext` turns a confusing bad run into an immediate error.
- **`scripts/bench/README.md:31,41`** — **(Correctness)** the README advertises the multi-project path ("8 for anything about drainers", `--tokens tokens.txt` round-robining) but neither `ramp.sh:42` nor `spike.sh:25` can use it — both pass a single `--token`. The multi-drainer measurement the README says the partitioned spool makes necessary is not reachable from any shipped driver. Either give the drivers a `--tokens` mode or say plainly that multi-project runs are `otlp-load.js` by hand.
- **`Taskfile.yml:243,644`** — **(Consistency)** this PR's diff carries the `evals-api-paid.json` -> `tessary-api-paid.json` rename from commit `ef9923f`, which is not on `main` and has no PR of its own. Merging #8 lands an unrelated rename under a `docs(bench)` title. Either rebase it out or say so in the PR body.

### Suggestions

- `scripts/bench/ramp.sh:34,56` — column 1 is headed `step` but holds the offered rate; column 2 is headed `offered_spans_per_s` but holds the *measured* rate. Rename to `offered` / `measured`.
- `scripts/bench/mem-probe.sh:20` — `-v "$PWD/$OUT:/out"` is mounted and never used. Drop it.
- `docker-compose.bench.yml:17` — the box table says "frontend/caddy"; there is no `caddy` service in `docker-compose.yml` (postgres, backend, frontend, redpanda behind the kafka profile, sandbox-runner + its init). Say `frontend`.
- `docker-compose.bench.yml:44` — `PYROSCOPE_AGENT_ENABLED: "false"` is a no-op: `devdocs/reference/config-keys.md:125` records that the open `docker-compose.yml` no longer wires the var at all. Harmless, but it reads as though it were doing something.
- `scripts/bench/otlp-load.js:180` — `req.on("error")` decrements `inflight` unconditionally; an error emitted after the response `end` handler already decremented double-decrements and slowly inflates apparent headroom over a long run. A per-request `settled` flag closes it.
- `scripts/bench/README.md:75-76` — worth noting that `IngestThroughputReporter` (`backend/substrate/src/main/java/ai/tessary/ingest/substrate/IngestThroughputReporter.java:74`) has `initialDelayString = "60000"`, so the first heartbeat is 60 s after backend boot whatever the cadence is. A short first ladder step will legitimately have no heartbeat lines, and that should not read as a finding.

Verdict: changes-requested.

scripts/bench/projects.sh
```diff
@@ -0,0 +1,36 @@
...
+#
+# The in-process spool partitions by project, so a single-project load test measures a single drainer
+# no matter how many are configured. This is what gives the drain something to parallelise.
+#
+#   bash projects.sh 8
+set -euo pipefail
+cd "$(dirname "${BASH_SOURCE[0]}")"
+N="${1:-8}"
+B=http://localhost:8000
+C="$(mktemp -t tessary-bench-cookies)"
```
nidhi-nair: **Blocker (Correctness)**: `mktemp -t <name>` without X's is a BSD/macOS-ism. On GNU coreutils this is `mktemp: too few X's in template`, exit 1 (verified in debian:12), so under `set -euo pipefail` the script aborts immediately on every Linux box, the self-host shape the README targets. Use `mktemp -t tessary-bench-cookies.XXXXXX`. Also worth a `trap 'rm -f "$C"' EXIT` so the session cookie jar is not left behind.

scripts/bench/projects.sh
```diff
@@ -0,0 +1,36 @@
...
+rm -f tokens.txt
```
nidhi-nair, line 14: **Blocker (Security)**: plaintext write-scoped bearer tokens land in `scripts/bench/tokens.txt`, inside the working tree, and nothing gitignores it (`.gitignore` covers `.local/` only). Same for the `results/` trees. `.gitleaks.toml:19` recognises `tsy_`, so the likely outcome is a tripped scanner, but if it is not caught it is a live token on `main`. Add `scripts/bench/tokens.txt` and `scripts/bench/results/` to `.gitignore` in this PR.

docker-compose.bench.yml
```diff
@@ -0,0 +1,73 @@
...
+      SPRING_APPLICATION_JSON: ${SPRING_APPLICATION_JSON:-}
+    volumes:
+      - ./.local/bench/jfr:/jfr
+      # The shipped image is a JRE: it has java and jfr but no jcmd, so there is no way to ask the
+      # live JVM for an NMT summary or a heap info. A sidecar JDK sharing this container's PID
+      # namespace can, but HotSpot's attach handshake goes through a socket in the target's /tmp,
+      # which a sidecar in its own mount namespace cannot see. Making /tmp a named volume both
+      # containers mount is what lets the handshake complete. Storage only; nothing about the
+      # application changes.
+      - bench-tmp:/tmp
```
nidhi-nair, line 59: **Warning (Security)**: the mechanism is well explained above, but not the consequences for someone who leaves the overlay on. (1) The HotSpot attach socket now sits in a volume any container mounting `tessary_bench-tmp` can reach, and the attach API includes `JVMTI.agent_load`, which is arbitrary code in the backend JVM; `mem-probe.sh` already demonstrates the attach as `--user root`. (2) A named volume is not wiped on restart, so whatever the app spools to `/tmp` accumulates on the host. Please add an explicit "for a measuring session only, take it back off afterwards" line here and in the README (that also covers leaving JFR + NMT enabled).

scripts/bench/README.md
```diff
@@ -0,0 +1,93 @@
...
+between arms so index size does not confound the result.
+
+```bash
+# Redaction parallelism (config-keys.md: 476 -> 1021 spans/s, 2.14x)
+SPRING_APPLICATION_JSON='{"tessary":{"redaction":{"parallelism":1}}}' \
+  docker compose -f docker-compose.yml -f docker-compose.bench.yml up -d --force-recreate backend
+node otlp-load.js --token "$(head -1 tokens.txt)" --rate 1200 --spans-per-request 40 --duration 90 --warmup 15
+# then the same with parallelism 4
+
+# Token cache (the p50 61 ms -> 2.6 ms control)
```
nidhi-nair: **Warning (Correctness)**: `61 ms -> 2.6 ms` is attributed to the `tessary.auth.token-cache.*` entry, but `devdocs/reference/config-keys.md:47` carries only "bcrypt was 62% of backend CPU on the ingest path" and the strings `61 ms` and `2.6 ms` appear nowhere in the tree. The redaction pair on line 51 does check out (config-keys.md:42, 476 -> 1,021, 2.14x); this one does not. Either land the number in config-keys.md in this PR or cite what is actually published.

scripts/bench/mem-probe.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "=== $(date -u +%FT%TZ)  tag=$TAG"
+  echo
+  echo "--- cgroup (the ceiling the JVM sized itself against)"
+  docker exec tessary-backend-1 sh -c 'echo "memory.max   $(cat /sys/fs/cgroup/memory.max)"
+    echo "memory.current $(cat /sys/fs/cgroup/memory.current)"
+    echo "memory.peak    $(cat /sys/fs/cgroup/memory.peak 2>/dev/null || echo n/a)"
+    grep -E "^(anon|file|slab|sock) " /sys/fs/cgroup/memory.stat
+    echo "cpu.max      $(cat /sys/fs/cgroup/cpu.max)"'
+  echo
+  echo "--- heap, after a full GC (live set, not floating garbage)"
```
nidhi-nair: **Warning (Correctness)**: `GC.heap_info` reports current occupancy and triggers no collection, so the recorded figure is not "after a full GC" and does include floating garbage. Either `jcmd_in GC.run` first or drop the parenthetical. This file exists to make the memory numbers trustworthy.

scripts/bench/ramp.sh
```diff
@@ -0,0 +1,68 @@
...
+  local out="$1" secs="$2"
+  local end=$((SECONDS + secs))
+  while [ "$SECONDS" -lt "$end" ]; do
+    docker stats --no-stream --format '{{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}' \
+      tessary-backend-1 tessary-postgres-1 2>/dev/null | sed "s/^/$(date +%s)\t/" >> "$out" || true
+    sleep 2
+  done
+}
+
+echo "step,offered_spans_per_s,sent,ok,refused_503,other,errors,rows_written,drain_spans_per_s,p50_ms,p95_ms,p99_ms" > "$OUTDIR/steps.csv"
```
nidhi-nair: **Warning (Correctness)**: `steps.csv` never records `skipped`. `otlp-load.js:152` drops a request without sending once `inflight >= --max-inflight` (512), which is exactly when the server is struggling, so the generator quietly stops being open-loop and the table cannot show it. Given README:22 makes open-loop the load-bearing claim, add `skipped` to the header and the row. Two naming nits while here: column 1 is headed `step` but holds the offered rate, and column 2 is headed `offered_spans_per_s` but holds `measured_span_rate`. And `accepted_span_rate`, which README:68 calls "the capacity figure", is never recorded at all.

PR #14, opened 2026-09-09
refactor(release): publish to Docker Hub only, and make a failed run clean up after itself

infinitetrooper:
Follow-up to run [34332530449](https://github.com/tessaryai/tessary/actions/runs/34332530449), which died merging to GHCR after Docker Hub had already taken half a release.

## GHCR is gone from the release path

It was published alongside Docker Hub on the theory that `GITHUB_TOKEN` with `packages: write` made a mirror free. It isn't: that token can only write a package **linked to the repository running the workflow**, and the `tessaryai/tessary` package — created outside this workflow — answers:

```
denied: permission_denied: read_package
```

A second registry that can fail independently of the first is a second way to publish half a release, and it bought nothing a self-hoster asked for: `docker pull org/image` resolves to Docker Hub by convention, which is where the one-command install already points. The namespace stays ours and stays in `namespace-inventory.txt`; nothing publishes to it.

Dropped with it: `GHCR_REPO`, `packages: write`, three `docker/login-action` steps, and the GHCR half of `publish-compose-artifact.sh`'s default repo list.

## Nothing irreversible happens until the release is proven

This is what makes "clean up on failure" actually possible.

`merge-and-tag` did three things at once: created `<service>-<version>`, repointed `<service>-latest`, and pushed the git tag. Two of those cannot be undone by deleting what the run made — removing a moved alias is worse than the move, and a pushed tag makes the version unreleasable again.

So it's split:

| job | writes |
|---|---|
| `merge-manifests` | `<service>-<version>` only |
| `verify-manifests` | — |
| `publish-compose` | `compose`, `compose-<version>` |
| **`finalize`** | the four `<service>-latest`, then the git tag `v<version>` |
| `github-release` | the Release |

Everything before `finalize` is additive, on tags nobody was using. The state this run left behind — `backend-latest` on 0.2.0 with three services carrying no `-latest` at all — is now unreachable.

## `cleanup` runs always, scoped to this run

Every tag it can name carries **this dispatch's own version**, so no earlier release is reachable from it.

- **Release succeeded** → the eight arch-suffixed scratch tags go.
- **Anything went red** → this run's four `<service>-<version>` manifests go too, so a half-finished release leaves nothing behind and the same version can be dispatched again cleanly.

`<service>-latest` is touched in neither mode and never needs to be — `finalize` is the only job that moves it, nothing runs after `finalize` but the announcement, and a run that died earlier never moved it.

Never fatal. It can't fail a release that succeeded, nor turn one failure into a confusing second one. Deleting a tag needs the Docker Hub API rather than the CLI, and a `DOCKER_TOKEN` scoped Read & Write but not Delete answers 403 — that earns a `::warning::`, not a red.

## `verify-registry-parity` → `verify-manifests`

With one registry there's no parity left to compare, but D3's reason for wanting a distinct job that *can* go red survives intact: "`merge-manifests` exited 0, so the tags are fine" is an assertion that never reddens, and a half-succeeded merge is exactly what this run produced.

It now asks the registry whether every tag resolves and whether each carries **both platforms**. An arch silently missing from a manifest list is invisible until a self-hoster on that arch tries to pull.

## Resulting shape

```
preflight
  → 8 build jobs
  → verify-open-artifacts (amd64 + arm64)
  → merge-manifests      <service>-<version>
  → verify-manifests     resolves + both platforms
  → publish-compose      compose, compose-<version>
  → finalize             -latest ×4, then git tag v<version>
  → github-release
  → cleanup              if: always(), this run's tags only
```

## Verification

- YAML parses; 12 jobs; `needs` graph as above; `permissions` is now `contents: write` alone; `env` is `DOCKER_REPO` alone.
- No stale references to `merge-and-tag`, `verify-registry-parity`, `cleanup-scratch-tags`, `GHCR_REPO`, or "both registries" anywhere in the tree. The one remaining `ghcr` mention in release.yml is the header paragraph explaining why it was removed.
- `check-required-inputs.sh`, `check-version-consistency.sh` and `check-compose-artifact.sh` all green; `bash -n` clean on every edited script.

**Not executed.**

## Before re-dispatching

Docker Hub still holds the wreckage of run 34332530449 and wants cleaning by hand — the new cleanup job only ever touches its own run's tags:

- `backend-0.2.0`, and `backend-latest` currently pointing at it
- `backend-0.2.0-{amd64,arm64}`, `frontend-0.2.0-{amd64,arm64}`, `sandbox-runner-0.2.0-{amd64,arm64}`, `agent-sandbox-0.2.0-{amd64,arm64}`
- the older `-alpha` set from an earlier rehearsal

`0.2.0` is still free — no `v0.2.0` tag was created, so preflight will accept it again.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -154,10 +154,9 @@ TESSARY_TELEMETRY_ENABLED=true
 # Which release this stack runs
 # ============================================================================
 #
-# TESSARY_VERSION moves all four images at once — one repo, per-service version tags, mirrored at
-# ghcr.io/tessaryai/tessary. Set one of the per-service keys directly to pin just that service to
-# something else (a different version, a local build tag). No personal namespace is ever named
-# here (D9).
+# TESSARY_VERSION moves all four images at once — one repo on Docker Hub, per-service version
+# tags. Set one of the per-service keys directly to pin just that service to something else (a
+# different version, a local build tag). No personal namespace is ever named here (D9).
 #
 # UNSET MEANS `-latest`. From a clone, leaving TESSARY_VERSION out tracks the newest release; a
 # `docker compose pull` then moves the stack forward. Write a version below to stop that. The
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,11 +9,18 @@
 # reusable-ecs-deploy.yml, which used to be the hosted product's own image builds, moved to
 # tessary-paid/ in the same PR that generalized this file — see that plan's divergence log).
 #
-# Registry: Docker Hub as primary (`tessaryai/tessary`), GHCR as a free mirror
-# (`ghcr.io/tessaryai/tessary`) — the repo-wide choice recorded in that same plan (prior-art
-# rationale: every comparable self-hosted OSS project checked ships its primary image on Docker
-# Hub, and `docker pull org/image` with no registry host resolves to Docker Hub by convention).
-# GHCR needs no extra secret: GITHUB_TOKEN with `packages: write` is enough for a same-repo push.
+# Registry: DOCKER HUB, and only Docker Hub (`tessaryai/tessary`) — every comparable self-hosted
+# OSS project checked ships its primary image there, and `docker pull org/image` with no registry
+# host resolves to Docker Hub by convention, so it is the one a self-hoster reaches for.
+#
+# GHCR WAS A MIRROR HERE AND IS GONE. `ghcr.io/tessaryai/tessary` was published alongside on the
+# theory that GITHUB_TOKEN with `packages: write` made it free. It is not free: that token can only
+# write a package LINKED to the repository running the workflow, and a package created any other
+# way answers `denied: permission_denied: read_package` — which is how run 34332530449 died, after
+# Docker Hub had already taken half a release. A second registry that can fail independently of the
+# first is a second way to publish half a release, and it bought nothing a self-hoster asked for.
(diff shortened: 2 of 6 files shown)


infinitetrooper:
Pushed a fix for a hole in the job order this PR introduced, found while drawing the graph out.

`publish-compose` ran **before** `finalize`, and `cleanup` deleted this run’s `<service>-<version>` manifests whenever the run did not complete. The compose artifact *names* those manifests. So a `finalize` failure — a blocked tag push, a blip repointing `-latest` — would have left the floating `compose` tag, the one the marketing site publishes, resolving to a config whose images `cleanup` had just deleted. Every one-command install broken until the next release, from a run that never claimed to have succeeded.

Two changes, both needed:

1. **`publish-compose` now runs after `finalize`.** Reaching `finalize` green is what makes `<service>-<version>` permanent, so it is the earliest point at which anything may take a lasting reference to those tags. `github-release` moves behind `publish-compose` with it, so the announcement still follows the artifact it announces.
2. **`cleanup` gates on `finalize`, not on the run’s overall result.** Once `finalize` is green the release is committed, so a later red keeps every version tag and leaves a coherent release to finish by hand. Only a run that never reached `finalize` has its version tags removed — and at that point nothing downstream exists to hold a reference.

Updated order:

```
merge-manifests → verify-manifests → finalize → publish-compose → github-release → cleanup
```

`check-required-inputs.sh`, `check-version-consistency.sh` and `check-compose-artifact.sh` all still green.

PR #72, opened 2026-09-16
feat(dev): task dev asks its questions, builds agents from a local image, and drops the local backend

infinitetrooper:
## Summary

Five fixes to the dev stack, building toward a checkout that comes up cleanly and behaves like a self-hosted install:

- **`fix(dev): mount tokens.css into the frontend dev container`** — the dev image's bind-mount list never gained `tokens.css` (frontend root, imported by `src/index.css`), so Vite served an error overlay instead of the app. The production Dockerfile already copies it; only dev was affected.
- **`fix(dev): dev:local no longer dies silently when there is no .env`** — the launcher-key lookup piped a `sed` over `.env` with stderr suppressed; with no `.env` (every fresh clone or worktree) `sed` exited non-zero, `pipefail` + `set -e` took the script down, and nothing printed why. The read is now guarded on the file existing.
- **`fix(dev): the dev stack ships working defaults for both sealing keys`** — `TESSARY_SECRET_KEY` and `TESSARY_AUTH_COOKIE_PASSWORD` defaulted to blank, so a checkout with no `.env` looked healthy and then 500'd on first sign-up. They now default to the same two placeholders `docker-compose.yml` already ships (valid 32-byte base64, publicly known, refused by `PlaceholderSecretGuard` on any install with a `SITE_DOMAIN`); an `.env` still wins.
- **`feat(dev): task dev asks its questions like docker compose`** — where triage/RCA agents run, whether the encoder classifier service runs, and whether sign-in is enforced were baked-in defaults that made the dev stack a different product from a self-hosted install (auth off, no agent launcher wired). `scripts/lib/dev-choices.sh` now asks each as multiple choice on first run and remembers the answers in `.local/dev-choices.env` (gitignored); `task dev:configure` re-asks them.
- **`feat(dev): build the agent-sandbox image from the checkout, drop the local backend`** — `task dev` now builds `AGENT_IMAGE` from this checkout's `sandbox-runner/agent-sandbox/` before the stack comes up, instead of relying on a published tag, so the docker sandbox backend always runs the agent scripts you have on disk, and a build failure stops the run at boot rather than surfacing as an opaque 502 on the first `/rca` or `/triage`. `docker` is now the chooser's recommended answer and `e2b` is relabelled as the published template. The `local` backend — a host child process with no container/VM isolation, `task dev:local` — is removed everywhere (`server.js`'s `runScriptLocally`/`childEnvFor`/the `SANDBOX_BACKEND=local` branch, the dev scripts, Taskfile, `docker-compose.dev.yml`, and docs): it bought nothing docker didn't already cover once docker builds from the checkout. `server.js`'s docker-container `Env` construction is pulled into a `containerEnvFor` seam so it stays unit-testable with no Docker daemon, and `docker-compose.dev.yml`'s `group_add` default is fixed from `999` to `0`, matching `docker-compose.yml`'s own default (Docker Desktop mounts the socket `root:root`, so `0` is correct for most first-time `task dev` runs on Mac/Windows).

## How to test

```bash
# JS unit tests (launcher + agent-sandbox), the base-URL parity check:
task sandbox-runner:check

# Version/image-tag consistency across the repo:
bash scripts/check-version-consistency.sh

# Shell lint on the touched dev scripts:
shellcheck scripts/dev.sh scripts/dev-up.sh scripts/lib/dev-choices.sh

# Compose renders (both profiles):
docker compose -f docker-compose.dev.yml config -q
COMPOSE_PROFILES=launcher docker compose -f docker-compose.dev.yml config -q

# End to end, from a clean checkout:
task dev              # answer the three questions (docker/off/off is a fast, keyless run);
                       # confirm the agent-sandbox image builds and the stack comes up healthy
task dev:configure     # re-answer and confirm .local/dev-choices.env updates
```


infinitetrooper:
Closing: these changes are merged into the triage-rework branch and will ship in that single PR.

